PR #1, opened 2015-06-12
Change Dockerfile: uwsgi + migrations

marcofucci:
- delete nginx in Docker
- delete supervisord in Docker
- uwsgi only running in Docker
- add docker_entrypoint.sh script as Docker entry point
- make the entry point migrate if env var passed in
- configure django settings to use appropriate env vars


marcofucci:
@r4vi @SteveMarshall core review?

r4vi:
Looks good to me

Dockerfile
```diff
@@ -5,35 +5,27 @@ RUN echo "Europe/London" > /etc/timezone  &&  dpkg-reconfigure -f noninteractive
...
-
 ADD ./conf/uwsgi /etc/uwsgi
 
-ADD ./conf/nginx/nginx.conf /etc/nginx/nginx.conf
-ADD ./conf/nginx/sites-enabled /etc/nginx/sites-enabled
-ADD ./conf/supervisor /etc/supervisor
-
 ADD ./requirements/ /app/requirements/
-RUN pip3 install -r requirements/dev.txt
+RUN pip3 install -r requirements/prod.txt
```
SteveMarshall: If we're using the same Dockerfile everywhere, can we just have one `requirements.txt`?

mtp_api/templates/core/index.html
```diff
@@ -6,7 +6,7 @@
 <div class="row">
   <div class="small-12 columns moj-wrapper">
     <h1>
-      Yo MoJ!&nbsp;
+      Yo MoJ API!&nbsp;
```
SteveMarshall: Great feature! :smile:

Dockerfile
r4vi: dev.txt is for local development

docker_entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
SteveMarshall: We already have a `./docker/`; maybe this could go in there as `run.sh`?

Dockerfile
SteveMarshall: Ah, right. OK then!

docker_entrypoint.sh
r4vi: `./docker` is really for supplementary docker images needed by docker-compose so maybe it's a bad name docker_entrypoint could go in `./bin/` if you don't like it toplevel

marcofucci: Well if we want to move it there, it would make sense to move the Dockerfile there as well as some other teams are doing.
The problem is that it might confuse people as they would expect a Dockerfile in the main folder.
The current docker folder is for dev only atm and I don't want to mix things up.

We should decide what we want/like and refactor it, in another PR maybe.

SteveMarshall: Both good arguments: let's leave it in the root, alongside the Dockerfile, named as-is. :+1:

PR #4, opened 2015-06-18
Adds auth to transaction API

r4vi:
This now limits the transactions a user can see
to their own prison. Unless they're a Django admin in
which case they can see everything.

Adds initial groups fixture for each user type but we haven't
mapped exactly what permission each group has, we want to do
this because each user type has a subset of permissions.

* We want `PrisonClerk` users to be able to mark
transactions as credited and see a list of them for their prison.
* We want `BankAdmin` to be able to upload a list of transactions
using the BAI2 file they download from online banking.
* We want `NOMSFinance` to be able to see all transactions, maybe run some reports.


mtp_api/apps/prison/models.py
```diff
@@ -7,5 +7,5 @@ class Prison(TimeStampedModel):
     nomis_id = models.CharField(max_length=3, primary_key=True)
     name = models.CharField(max_length=500)
 
-    def __unicode__(self):
+    def __str__(self):
```
SteveMarshall: Why `__str__()` instead of `__unicode__()`? I thought the latter was preferred these days?

SteveMarshall: Oh, is it because Python 3?

r4vi: yep

SteveMarshall:
It'd be nice to see some tests checking for various auth conditions (can I get a list of transactions for my prison, can I *not* get transactions for another prison) in here, I think?

r4vi:
now at -> https://github.com/ministryofjustice/money-to-prisoners-api/pull/7

PR #7, opened 2015-06-19
Feature/transactions auth

r4vi:
This now limits the transactions a user can see
to their own prison. Unless they're a Django admin in
which case they can see everything.

Adds initial groups fixture for each user type but we haven't
mapped exactly what permission each group has, we want to do
this because each user type has a subset of permissions.

* We want PrisonClerk users to be able to mark transactions as credited and see a list of them for their prison.
* We want BankAdmin to be able to upload a list of transactions using the BAI2 file they download from online banking.
* We want NOMSFinance to be able to see all transactions, maybe run some reports.


r4vi:
@marcofucci and @SteveMarshall can you review please?

README.md
```diff
@@ -36,3 +37,64 @@ you should be able to point your browser at
...
+```
+{
+    "count": 6,
+    "next": null,
+    "previous": null,
+    "results": [
+        {
+            "id": 240,
+            "prisoner_number": "R5957RF",
+            "prisoner_name": "Lewis Worthington",
```
marcofucci: delete prisoner_number

requirements/dev.txt
```diff
@@ -2,3 +2,5 @@
 -r base.txt
 django-extensions==1.5.0
 Werkzeug==0.10.4
+names==0.3.0
```
marcofucci: delete `names`, we are not using it anymore

mtp_api/apps/transaction/views.py
```diff
@@ -32,8 +34,24 @@ class Meta:
...
+
+    def get_prison_set_for_user(self):
+        try:
+            return PrisonUserMapping.objects.get(user=self.request.user).prisons.all()
+        except PrisonUserMapping.DoesNotExist:
+            return Prison.objects.none()
+
+    def get_queryset(self):
+        qs = super(OwnPrisonListModelMixin, self).get_queryset()
+        if self.request.user.is_superuser:
```
marcofucci: I read the TODO comment on the related test and I agree. Maybe it's worth deleting this `is_superuser` check and add it later on. Otherwise we might forget to add a testcase for this.

r4vi: agreed
This isn't really needed as nobody will be is_superuser anyway.

mtp_api/apps/transaction/tests/test_views.py
```diff
@@ -28,26 +39,57 @@ def assertTransactionsInUpload(self, response, upload_counter):
...
+
+        return Counter([x.prison for x in self.transactions if x.upload_counter == upload_counter])
+
     def test_list_latest_upload(self):
         url = reverse('transaction-list')
+        unique_prisons = self.calculate_prison_transaction_counts()
+        for prison, count in unique_prisons.items():
+            # we have to do if Prison because some transactions don't have a prison attached
+            # and therefore no user will be able to see those...
+            # TODO: figure out who is responsible for these transactions
```
SteveMarshall: Ultimately, those transactions will be refunded by the `BankAdmin`s; at some point in a future sprint, we need to give them a UI to download a file (in a format we don't yet know, of course!) to upload to our bank account to issue the money back to the sender. It may be that we make a product decision *not* to do that, mind you, but that remains to be seen.

SteveMarshall:
Is it worth explicitly testing that an authenticated user can't access transactions uploaded for a prison that's not their own when they're limited to a given prison?

r4vi:
@SteveMarshall how? there is no detail endpoint yet... so anyone going to `/transactions/1234` will just get a 404. You can currently only hit `/transactions/` and check that all of those transactions belong to you, you can't really do anything else. I'll add a test for that all the ones in the list belonging to the user but that's all we can do

SteveMarshall:
@r4vi Good point. It's been a long day :sweat:

r4vi:
@marcofucci / @SteveMarshall good to merge?

marcofucci:
:+1: for me

PR #8, opened 2015-07-03
Provide Makefile for convenience

matharden:
This makes it a lot easier to run the server everyday.


SteveMarshall:
This looks good. You should mark `build` and `run` as [`PHONY` targets](https://www.gnu.org/software/make/manual/make.html#Phony-Targets), though, so they don't break if we have files/folders with those names.

PR #9, opened 2015-07-03
Feature/accounting api 

r4vi:
Create an API for: 
  * marking transactions as credited
  * getting a list of transactions you've locked
  * getting a list of transactions for a prison
  * filter a list of transactions by status
  * lock / unlock a batch of transactions

finishes [#96083016] and [#96084532]


r4vi:
@SteveMarshall wanna review?

mtp_api/apps/transaction/tests/test_views.py
```diff
@@ -7,105 +7,744 @@
...
+        )
+
+
+class TransactionsEndpointTestCase(BaseTransactionViewTestCase):
+
+    def test_cant_access(self):
         """
-        Asserts that all the transactions in the response belong to
-        the specific upload batch
+        GET on transactions endpoint should 404.
```
SteveMarshall: This says 404, it says 403 below. Update this?

r4vi: yep should say 403; we changed our mind

mtp_api/apps/transaction/tests/test_views.py
```diff
@@ -7,105 +7,744 @@
...
+        response = self.client.get(url, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_all(self):
+        """
+        GET without params should return all transactions linked
+        to that prison.
+        """
+        self._request_and_assert()
```
SteveMarshall: I'm not *entirely* keen on this pattern, but I can see why it appeals. Just might be a bit nicer if the tests were a bit more explicit in what they were doing, perhaps?

mtp_api/apps/transaction/tests/test_views.py
```diff
@@ -7,105 +7,744 @@
...
+
+        url = self._get_url(transactions_owner, prison, 1)
+        response = self.client.post(url, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+        # check 5 taken transactions in db
+        self.assertEqual(
+            self._get_pending_transactions_qs(prison, transactions_owner).count(),
+            0
```
SteveMarshall: Should this be asserting against something other than 0? (The comment suggests it should be 5?)

mtp_api/apps/transaction/tests/test_views.py
```diff
@@ -7,105 +7,744 @@
...
+        # check TAKE_LIMIT-1 taken transactions in db
+        self.assertEqual(
+            self._get_pending_transactions_qs(prison, user).count(), count
+        )
+
+        # request 2 more => should fail
+        url = self._get_url(user, prison, 2)
+        response = self.client.post(url, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
SteveMarshall: Might make sense for this to be `409 Conflict`, maybe?

mtp_api/apps/transaction/views.py
```diff
@@ -1,37 +1,39 @@
...
-            # get the latest if 'upload_counter' param not specified
-            try:
-                transaction = Transaction.objects.latest('upload_counter')
-                upload_counter = transaction.upload_counter
-            except Transaction.DoesNotExist:
-                pass
-        return upload_counter
+    def filter(self, qs, value):
+
+        if value in ([], (), {}, None, ''):
```
SteveMarshall: Any reason this isn't `if not value`?

mtp_api/apps/transaction/tests/test_views.py
marcofucci: I see what you mean, 409 plus a message saying: "If you want to take that many transactions you need to release some others".
The 400 looks OK as well though as you don't really care *which* transactions you have taken, you only care about how many you took.

Not sure which one we should use, @r4vi @SteveMarshall ?

mtp_api/apps/transaction/views.py
```diff
@@ -52,9 +54,87 @@ class TransactionView(
...
+
+        if prison_id and filter_by_prison:
+            qs = qs.filter(prison_id=prison_id)
+
+        if user_id and filter_by_user:
+            qs = qs.filter(owner__id=user_id)
+
+        return qs
+
+    def take(self, request, *args, **kwargs):
```
SteveMarshall: The more I think about it, the more I think this should be `assign` rather than `take`. Waddya think?

mtp_api/apps/transaction/tests/test_views.py
marcofucci: @SteveMarshall Yes, I started like that but they were getting very long, repetitive and complicated so I thought this looked better.
Any suggestions on how to improve it? Otherwise, I'd rather keep it like this.

mtp_api/apps/transaction/views.py
r4vi: that breaks on 0

mtp_api/apps/transaction/views.py
SteveMarshall: That said, in the context of a user taking them to work on, `take` makes more sense. Hmmmm.

mtp_api/apps/transaction/views.py
SteveMarshall: Oh, yeah, of course it does. `False` too.

mtp_api/apps/transaction/views.py
marcofucci: yeah, I think *take* is better as you take a transaction, you don't *assign* it to yourself.

mtp_api/apps/transaction/tests/test_views.py
SteveMarshall: Not sure right now. Maybe we should have a quick look together on Monday morning?

mtp_api/apps/transaction/tests/test_views.py
SteveMarshall: I'm happy enough with it being as it is, but I think your suggestion of 409 and a suggestion to release some would be nice?

mtp_api/apps/transaction/views.py
```diff
@@ -52,9 +54,87 @@ class TransactionView(
...
+        self.check_permissions(request)
+
+        slice_size = self.get_slice_limit(request)
+
+        with transaction.atomic():
+            pending = self.get_queryset(filter_by_user=False).available().select_for_update()
+            slice_pks = pending.values_list('pk', flat=True)[:slice_size]
+
+            queryset = self.get_queryset(filter_by_user=False).filter(pk__in=slice_pks)
+            queryset.update(owner=request.user)
```
SteveMarshall: Doesn't this clash with the route? If I understand correctly, the route is `<prison_id>/<user_id>/take'`, but we're passing `request.user`… Doesn't that mean I could hit `<prison_id>/<someone_else>/take` and it would assign to me? What's the expected behaviour? If we're just using `request.user`, I'd suggest we make the route `<prison_id>/take`.

r4vi: https://github.com/ministryofjustice/money-to-prisoners-api/pull/9#discussion-diff-33869619R83 stops them from accessing that route if they're not the same  - we could just as easily use kwargs['user_id'] instead but this saves a db lookup

SteveMarshall:
Where we have `self.assertEqual(response.status_code, status.HTTP_303_SEE_OTHER)`, is it worth checking the response's `Location` header to ensure it's directing the client somewhere sensible?

mtp_api/apps/transaction/views.py
SteveMarshall: OK!

PR #11, opened 2015-07-09
Transaction logs

marcofucci:
Logs created when a transaction gets created / taken / released / credited.
The records contain info about who did what.

This is a high level logging solution meant to be used mainly by admins to check the activities on a transaction. It is not meant to be used for security purposes / intrusion detection and it does not contain technical / specific details about requests.


marcofucci:
@SteveMarshall this is how I would do this + server logs to identify hostile activities as they are different things. What do you think?
With this, we will not be able to store IP details etc. (if we need to in the future) as this is an app logging layer.

marcofucci:
@SteveMarshall can we merge this?

PR #12, opened 2015-07-13
PrisonerLocation model and endpoint

marcofucci:
New endpoint allowing the creation of multiple prisoner locations at the same time.
This simulates a report upload but it's a lot cleaner/easier to understand and safer.

The format is:

    CREATE   /prisoner-locations/
        [
        	{
        		prisoner_number: "<prisoner-number>",  # e.g. L2513UV
        		prisoner_dob: "<prisoner-dob>",  # e.g. 1980-06-17
        		prison: "<prison-id>"  # e.g. prison-1
        	},
        	{
        		prisoner_number: "<prisoner-number>",  # e.g. L2513UV
        		prisoner_dob: "<prisoner-dob>",  # e.g. 1980-06-17
        		prison: "<prison-id>"  # e.g. prison-1
        	}
        ]  

The system would store the items and update an `upload_counter` incremented field on each one of them which can be used later on to group them.

**Pros**
 * easy to understand
 * we don't need to store any files

**Cons**
 * clients would have to pre-process the data
 * each call to the endpoint represents an "upload" so partial creates are not allowed (2 calls => 2 uploads with the last one meant to "override" the first one).

### DIFF
--- a/mtp_api/apps/core/tests/utils.py
+++ b/mtp_api/apps/core/tests/utils.py
@@ -26,10 +26,10 @@ def make_test_oauth_applications():
     )
 
     Application.objects.get_or_create(
-        client_id='noms-admin',
+        client_id='prisoner-location-admin',
         client_type='confidential',
         authorization_grant_type='password',
-        client_secret='noms-admin',
-        name='noms-admin',
+        client_secret='prisoner-location-admin',
+        name='prisoner-location-admin',
         user=User.objects.first()
     )
--- a/mtp_api/apps/core/tests/utils.py
+++ b/mtp_api/apps/core/tests/utils.py
@@ -1,7 +1,9 @@
-from django.contrib.auth.models import User
-from mtp_auth.tests.mommy_recipes import create_prison_user_mapping
 from oauth2_provider.models import Application
+
+from django.contrib.auth.models import User
+
 from prison.models import Prison
+from mtp_auth.tests.mommy_recipes import create_prison_user_mapping
 
 
 def make_test_users(users_per_prison=1):
@@ -20,4 +22,14 @@ def make_test_oauth_applications():
         authorization_grant_type='password',
         client_secret='cashbook',
         name='cashbook',
-        user=User.objects.first())
+        user=User.objects.first()
+    )
+
+    Application.objects.get_or_create(
+        client_id='noms-admin',
+        client_type='confidential',
+        authorization_grant_type='password',
+        client_secret='noms-admin',
+        name='noms-admin',
+        user=User.objects.first()
+    )
--- a/mtp_api/apps/prison/models.py
+++ b/mtp_api/apps/prison/models.py
@@ -19,10 +19,3 @@ class PrisonerLocation(TimeStampedModel):
     prisoner_dob = models.DateField()
 
     prison = models.ForeignKey(Prison)
-
-    def __str__(self):
-        return 'location record ' \
-            'for prisoner {prisoner_number} / {prisoner_dob}'.format(
-                prisoner_number=self.prisoner_number,
-                prisoner_dob=self.prisoner_dob
-            )
--- a/mtp_api/apps/prison/models.py
+++ b/mtp_api/apps/prison/models.py
@@ -13,7 +13,6 @@ class Prison(TimeStampedModel):
 
 
 class PrisonerLocation(TimeStampedModel):
-    upload_counter = models.PositiveIntegerField()
     created_by = models.ForeignKey(settings.AUTH_USER_MODEL)
 
     prisoner_number = models.CharField(max_length=250)
@@ -22,9 +21,8 @@ class PrisonerLocation(TimeStampedModel):
     prison = models.ForeignKey(Prison)
 
     def __str__(self):
-        return 'location record {upload_counter} ' \
+        return 'location record ' \
             'for prisoner {prisoner_number} / {prisoner_dob}'.format(
-                upload_counter=self.upload_counter,
                 prisoner_number=self.prisoner_number,
                 prisoner_dob=self.prisoner_dob
             )
--- a/mtp_api/apps/prison/models.py
+++ b/mtp_api/apps/prison/models.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.db import models
 
 from model_utils.models import TimeStampedModel
@@ -9,3 +10,21 @@ class Prison(TimeStampedModel):
 
     def __str__(self):
         return self.name
+
+
+class PrisonerLocation(TimeStampedModel):
+    upload_counter = models.PositiveIntegerField()
+    created_by = models.ForeignKey(settings.AUTH_USER_MODEL)
+
+    prisoner_number = models.CharField(max_length=250)
+    prisoner_dob = models.DateField()
+
+    prison = models.ForeignKey(Prison)
+
+    def __str__(self):
+        return 'location record {upload_counter} ' \
+            'for prisoner {prisoner_number} / {prisoner_dob}'.format(
+                upload_counter=self.upload_counter,
+                prisoner_number=self.prisoner_number,
+                prisoner_dob=self.prisoner_dob
+            )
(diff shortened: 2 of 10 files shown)


marcofucci:
@SteveMarshall I'm not sure I like this but I'd like to know what you think about it.
The other option is to just upload the file and parse it (which to be honest I might like less).

SteveMarshall:
@marcofucci This is pretty much what I had in mind for the API. Other things I was considering were:

- If we need to be able to verify the uploads… Though I suppose we could do something like POSTing to `/prisoner-locations/[upload-id]/verify` or something? Or maybe consider uploads on the same day verification uploads? (But that's a feature for later)
- Whether we should actually encrypt them before upload, so the call would be:

```
CREATE   /prisoner-locations/
    [
        "encrypted_prisoner_number_dob_and_prison",
        "encrypted_prisoner_number_dob_and_prison",
    ]
```

That way, our systems never see or store the raw location/prisoner mapping, but I'm not sure if that's *really* worthwhile.

@ohmygauss: do you have any thoughts on if it's worth encrypting this data before it leaves Quantum? We'd be uploading ~85k lines per CREATE when going live.

ohmygauss:
1. Can also verify using a checksum, perhaps upload-id.timestamp? 

2. Data should be encrypted if its in transit if we're considering Quantum to be a blackbox. I'd imagine TLS >=1.1, strong ciphersuites and secure endpoints.

SteveMarshall:
@ohmygauss As discussed on Hipchat, let's talk about this when I'm in the office tomorrow.

SteveMarshall:
@ohmygauss and I concluded that we should upload hashes of the three fields, combined, so we have the set of all valid `prisoner ID + DOB + prison ID` hashes, then intersect this, when requesting transactions, with the set of all hashes of normalised transaction references (`prisoner ID + DOB`) and the `current prison ID`. That way, we're never storing the prisoners' locations in the clear.

We may need to pre-cache the transaction hashes (with the cache being time-bound), depending on the complexity of the algorithm we use, but we should investigate that further once we have this working.

mtp_api/apps/prison/serializers.py
```diff
@@ -0,0 +1,34 @@
...
+class PrisonerLocationListSerializer(serializers.ListSerializer):
+
+    def get_latest_upload_counter(self):
+        result = PrisonerLocation.objects.all().aggregate(max=Max('upload_counter'))
+        return result['max'] or 0
+
+    def create(self, validated_data):
+        # NOTE: this can cause race conditions and can be fixed but do we care?
+        # Not likely to happen
+        next_upload_counter = self.get_latest_upload_counter() + 1
```
SteveMarshall: Do we need the upload counter? Can we just use the locations that've been uploaded in the last day (or the most recent date we have locations for)?

mtp_api/apps/prison/tests/test_views.py
```diff
@@ -0,0 +1,124 @@
...
+                'data': [
+                    {
+                        'prisoner_number': random_prisoner_number(),
+                        'prisoner_dob': random_prisoner_dob(),
+                        'prison': 'invalid'
+                    }
+                ],
+                'msg': 'Should fail because invalid prison'
+            },
+        ]
```
SteveMarshall: I'm not particularly keen on this pattern: this one test could fail for a number of reasons around invalid data. Can we break this into multiple tests, each testing a single failure mode?

mtp_api/apps/prison/serializers.py
marcofucci: Yes we can but what if one uploads twice in a day? We would have to override the first upload which is not great as it might hide some issues. It would be better to always store every upload and get the latest records by MAX(upload_conter).

mtp_api/apps/prison/tests/test_views.py
marcofucci: To be honest, these validation checks are not very useful as provided by DSF by default which is tested on its own. I included it just in case we are doing something very stupid and the validation is bypassed.

If you're happy with it we could just remove this test, also because I'm testing just a subset of all possible scenarios.

marcofucci:
@ibrechin it looks like you finished this, right? If so @SteveMarshall can review it and we might be able to merge it?

mtp_api/apps/prison/tests/test_views.py
```diff
@@ -0,0 +1,133 @@
...
+            },
+            {
+                'data': [
+                    {
+                        'prisoner_number': '*'*1000,
+                        'prisoner_dob': random_prisoner_dob(),
+                        'prison': self.prisons[0].pk
+                    }
+                ],
+                'msg': 'Should fail because empty data'
```
SteveMarshall: Is this message right, or is this a C&P error?

marcofucci: oh copy/paste eheh, btw what did we decide to do with these?

mtp_api/apps/prison/tests/test_views.py
SteveMarshall: I still think this should be a bunch of individual tests (maybe using a [data provider](https://pypi.python.org/pypi/unittest-data-provider) or similar), but I'm not bothered enough that it should block us merging.

mtp_api/apps/prison/views.py
```diff
@@ -1 +1,28 @@
...
+from .models import PrisonerLocation
+from .serializers import PrisonerLocationSerializer
+
+
+class PrisonerLocationView(
+    mixins.CreateModelMixin, viewsets.GenericViewSet,
+):
+    queryset = PrisonerLocation.objects.all()
+
+    permission_classes = (IsAuthenticated, )
```
SteveMarshall: I assume we'll want to add to this once #17 is merged.

mtp_api/apps/core/tests/utils.py
```diff
@@ -20,4 +22,14 @@ def make_test_oauth_applications():
         authorization_grant_type='password',
         client_secret='cashbook',
         name='cashbook',
-        user=User.objects.first())
+        user=User.objects.first()
+    )
+
+    Application.objects.get_or_create(
+        client_id='noms-admin',
```
marcofucci: @SteveMarshall not sure about the name `noms-admin` looks rubbish, what do you think?

mtp_api/apps/prison/models.py
```diff
@@ -9,3 +10,19 @@ class Prison(TimeStampedModel):
...
+    prisoner_dob = models.DateField()
+
+    prison = models.ForeignKey(Prison)
+
+    def __str__(self):
+        return 'location record ' \
+            'for prisoner {prisoner_number} / {prisoner_dob}'.format(
+                prisoner_number=self.prisoner_number,
+                prisoner_dob=self.prisoner_dob
+            )
```
SteveMarshall: When're we likely to use this? I wonder if it's worth not having (or having a method that explicitly doesn't return prisoner details), in case we end up leaking prisoner locations into our logs?

mtp_api/apps/prison/tests/test_views.py
```diff
@@ -0,0 +1,133 @@
...
+        response = self.client.post(
+            self.list_url, data=data, format='json'
+        )
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        # test that total prisoner location records is now 3
+        latest_created = PrisonerLocation.objects.all()
+        self.assertEqual(latest_created.count(), len(data))
+        for item in data:
+            self.assertEqual(latest_created.filter(**item).count(), 1)
```
SteveMarshall: I'd like to see this in a separate `test_overwrite`: that way, we can see when we've only broken, and when we've only broken overwriting.

SteveMarshall: (As with the `test_create_validation_error` stuff, though, I'm not averse enough to this that it should block this being merged, if you both think it's alright.)

mtp_api/apps/core/tests/utils.py
SteveMarshall: Yeah… And that's not exactly what the role is, either. `prisoner-location-admin`?

marcofucci:
@SteveMarshall I've covered all the points, can you please check and merge?

SteveMarshall:
:+1: :sparkles:

PR #14, opened 2015-07-28
Add permission checks to transaction endpoints

marcofucci:
So that only users with the right permissions can access the transaction endpoints.
Done by using Django permissions + DRF Django Permission class.
I had to subclass the latter to support permission checks on GETs as not enabled by default.

We will use Django groups (Prison Clerk in this case) to manage these permissions.

### DIFF
--- a/mtp_api/apps/transaction/tests/test_views.py
+++ b/mtp_api/apps/transaction/tests/test_views.py
@@ -27,7 +27,10 @@ def get_prisons_for_user(user):
 
 
 class BaseTransactionViewTestCase(APITestCase):
-    fixtures = ['test_prisons.json']
+    fixtures = [
+        'initial_groups.json',
+        'test_prisons.json'
+    ]
     STATUS_FILTERS = {
         None: lambda t: True,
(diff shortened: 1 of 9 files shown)


marcofucci:
@SteveMarshall review/merge ?

mtp_api/apps/transaction/tests/test_views.py
```diff
@@ -143,6 +146,25 @@ def test_fails_with_logged_in_managing_different_prison(self):
...
+        # delete user from groups
+        user.groups.all().delete()
+
+        url = self._get_list_url(prison)
+
+        self.client.force_authenticate(user=user)
+
+        response = self.client.get(url, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
```
SteveMarshall: As discussed, it'd be nice to reduce some of the duplication, maybe with a mixin or something.

marcofucci:
Renamed to `TransactionRejectsRequestsWithoutPermissionTestMixin` as you suggested.
It's just a name and it's only used in that file so I'm OK with it being ugly.

SteveMarshall:
:+1:

PR #15, opened 2015-07-28
Add createdb instructions to README

marcofucci:
I've tried to find an easy / clear way of doing this using docker-compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,20 @@ and if you're planning to deploy then you'll need the [deployment repository](ht
 
 In a terminal `cd` into the directory you checked this project out into, then
 ```
-$ docker-compose build && docker-compose up
+$ docker-compose build
+```
+Create the database:
+```
+$ docker-compose run db /bin/bash
+$ ./docker-entrypoint.sh postgres &
+$ su postgres
+$ createdb mtp_api;
+$ exit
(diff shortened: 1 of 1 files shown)


marcofucci:
@SteveMarshall what do you think?

SteveMarshall:
Is there a reason we can't just set the `POSTGRES_DB` environment variable and [let the entry point set up the `mtp_api` DB](https://github.com/docker-library/postgres/blob/master/9.4/docker-entrypoint.sh#L64)?

If that doesn't work, I can't see any reason not to use the default `postgres` DB. As the docs say, it's “meant for use by users, utilities and third party applications”, and this is only for local development (and maybe testing on Jenkins), so we should just be able to make `mtp_api/settings/base.py` call `os.environ.get('DB_NAME', 'mtp_api')`, and pass that in the `docker-compose.yml` like we do for host, port, and so on.

marcofucci:
@SteveMarshall that might work but you need to delete the postgres dbdata image and test it as the docker_entrypoint.sh really executes only ones.

I will leave this to you as it's not high priority for now and I'd rather focus on other things.

SteveMarshall:
I've tackled using the default postgres container DB in #16, so closing this.

PR #16, opened 2015-07-29
Make DB name configurable for use in devenvs

SteveMarshall:
At present, we have to manually initialise the containers' `mtp_api` database, but the postgres image we're using for our database comes with a pre-configured `postgres` database.

[The documentation](https://registry.hub.docker.com/_/postgres/) says this database is “meant for use by users, utilities and third party applications”. Given that docker-compose is only used for development/testing, we probably don't need a dedicated new DB created, so we can just use the default postgres DB in Django.

I also tried setting a `POSTGRES_DB` environment variable and [letting the postgres container's entrypoint set up the database](https://github.com/docker-library/postgres/blob/master/9.4/docker-entrypoint.sh#L64), but the new DB doesn't appear to be ready in time for the django container to run migrations et al.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,7 @@ django:
   links:
     - db
   environment:
+    DB_NAME: postgres
     DB_USERNAME: postgres
     DB_PASSWORD: postgres
     DB_HOST: db_1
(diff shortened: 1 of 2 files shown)


marcofucci:
Using `postgres` as db name is probably very confusing and I'd rather try and use `mtp_api` instead but if it's a pain, let's keep it this way.

SteveMarshall:
Yeah, I'm not keen on having `postgres` as the DB, but it's only in `docker-compose` environments and, as I said in the original comment, I tried to automate the creation of `mtp_api` DB, but it doesn't get created soon enough in the postgres entrypoint for Django to find it.

I think a tiny reduction in clarity of local DB names is worth it if it makes getting started quicker and more automated.

PR #17, opened 2015-08-07
Refactor permission checking

marcofucci:
This creates a new ActionsBasedPermissions to manage permissions on a view.

The old version couldn't distinguish between a POST to create a transaction
and a POST to take a transaction.

The new version checks for individual permissions so that they can be
assigned and managed better.

### DIFF
--- a/mtp_api/apps/transaction/permissions.py
+++ b/mtp_api/apps/transaction/permissions.py
@@ -16,14 +16,11 @@ class IsOwnPrison(BasePermission):
 
 
 class TransactionPermissions(ActionsBasedPermissions):
-    actions_perms_map = {
-        'create': ['%(app_label)s.add_%(model_name)s'],
+    actions_perms_map = ActionsBasedPermissions.actions_perms_map.copy()
+    actions_perms_map.update({
         'list': ['%(app_label)s.view_%(model_name)s'],
         'retrieve': ['%(app_label)s.view_%(model_name)s'],
-        'update': ['%(app_label)s.change_%(model_name)s'],
-        'partial_update': ['%(app_label)s.change_%(model_name)s'],
-        'destroy': ['%(app_label)s.delete_%(model_name)s'],
         'take': ['%(app_label)s.take_%(model_name)s'],
         'release': ['%(app_label)s.release_%(model_name)s'],
         'patch_credited': ['%(app_label)s.patch_credited_%(model_name)s']
-    }
+    })
--- a/mtp_api/apps/transaction/permissions.py
+++ b/mtp_api/apps/transaction/permissions.py
@@ -1,11 +1,29 @@
 from rest_framework.permissions import BasePermission
 
+from core.permissions import ActionsBasedPermissions
+
+
 class IsOwner(BasePermission):
 
     def has_permission(self, request, view):
         return str(request.user.id) == view.kwargs.get('user_id')
 
+
 class IsOwnPrison(BasePermission):
 
     def has_permission(self, request, view):
         return request.user.prisonusermapping.prisons.filter(pk=view.kwargs.get('prison_id')).exists()
+
+
+class TransactionPermissions(ActionsBasedPermissions):
+    actions_perms_map = {
+        'create': ['%(app_label)s.add_%(model_name)s'],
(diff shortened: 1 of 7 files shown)


marcofucci:
@SteveMarshall @ibrechin review/merge?

mtp_api/apps/transaction/permissions.py
```diff
@@ -1,11 +1,29 @@
...
+        'create': ['%(app_label)s.add_%(model_name)s'],
+        'list': ['%(app_label)s.view_%(model_name)s'],
+        'retrieve': ['%(app_label)s.view_%(model_name)s'],
+        'update': ['%(app_label)s.change_%(model_name)s'],
+        'partial_update': ['%(app_label)s.change_%(model_name)s'],
+        'destroy': ['%(app_label)s.delete_%(model_name)s'],
+        'take': ['%(app_label)s.take_%(model_name)s'],
+        'release': ['%(app_label)s.release_%(model_name)s'],
+        'patch_credited': ['%(app_label)s.patch_credited_%(model_name)s']
+    }
```
ibrechin: Possibly update the superclass dict rather than override it? Just in the sense that it avoids duplicating default permissions i.e.

```python
class TransactionPermissions(ActionsBasedPermissions):
    ActionsBasedPermissions.action_perms_map.update({
       'destroy': ['%(app_label)s.delete_%(model_name)s'],
       'take': ['%(app_label)s.take_%(model_name)s'],
       'release': ['%(app_label)s.release_%(model_name)s'],
       'patch_credited': ['%(app_label)s.patch_credited_%(model_name)s']
    })
```

marcofucci: `update` updates the dict which is mutable so it would update `ActionsBasedPermissions.action_perms_map` as well.

SteveMarshall: Yeah, I think that also makes it more explicit when we want to use default permissions and add some versus entirely overriding the defaults.

SteveMarshall: @marcofucci I guess we'd have to `.copy().update()` (which is less clean), but I think it's worth it for the increased clarity?

SteveMarshall:
Should we have some tests for `ActionsBasedPermissions`?

mtp_api/apps/transaction/permissions.py
ibrechin: Could also do it in an ```__init__``` override.

```python
        def __init__(self, *args, **kwargs):
                self.action_perms_map.update({...})
                super(ActionBasedPermissions, self).__init__(*args, **kwargs)
```

That feels a bit dodgy, though, and could produce unintended behaviour if e.g. someone subclasses a subclass with an update in the init method and then also redefines the attribute in an attempt to override completely.

SteveMarshall:
@marcofucci Also: first line of commit should be “Refactor […]”, not “Refactored” :wink:

mtp_api/apps/transaction/permissions.py
marcofucci: `update` doesn't return anything so it won't work

ibrechin: @marcofucci can do:

```python
p = ActionBasedPermissions.action_perms_map.copy()
p.update({...})
```

marcofucci: Didn't know that you guys would have hated overriding the dict so much :-|
I don't like so much the other options but I'll go for whatever you guys like, it's a small thing so I'm not picky.

@SteveMarshall are you happy with 

```
class TransactionPermissions(ActionsBasedPermissions):
    action_perms_map = ActionBasedPermissions.action_perms_map.copy()
    action_perms_map.update({...})
```
?

SteveMarshall: @marcofucci I don't hate it *that* much, I was just trying to work out if there was a nice solution. I think, of all of them, that solution (`.copy()` then `.update()`) is the neatest with the least duplication.

marcofucci:
@SteveMarshall I'll be more careful next time ;-)

marcofucci:
@SteveMarshall I think everything is now covered, can you merge it if happy?

SteveMarshall:
:boom:

PR #19, opened 2015-08-11
Prisoner hash

marcofucci:
This adds a new field `prisoner_hash` to `PrisonerLocation` so that we can join tables easily using this value instead of `(prisoner_id, prisoner_dob)`.

I've also cleaned up the imports making sure we always use `get_user_model` instead of the `auth.User` one.

### DIFF
--- a/mtp_api/apps/prison/migrations/0005_populate_prisoner_hash.py
+++ b/mtp_api/apps/prison/migrations/0005_populate_prisoner_hash.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+import hashlib
+
+from django.db import models, migrations
+
+
+def calculate_prisoner_hash(location):
+    original = '{number}_{dob}'.format(
+        number=location.prisoner_number.lower(),
+        dob=location.prisoner_dob.strftime('%d/%m/%Y')
+    )
+    hash_object = hashlib.sha256(original.encode())
+    return hash_object.hexdigest()
+
+
+def populate_prisoner_hash(apps, schema_editor):
+    PrisonerLocation = apps.get_model("prison", "PrisonerLocation")
+
+    for location in PrisonerLocation.objects.all():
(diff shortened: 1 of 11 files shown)


marcofucci:
@SteveMarshall @ibrechin sorry another one for you guys

SteveMarshall:
retest this please

mtp_api/apps/prison/migrations/0005_populate_prisoner_hash.py
```diff
@@ -0,0 +1,33 @@
...
+from django.db import models, migrations
+
+
+def calculate_prisoner_hash(location):
+    original = '{number}_{dob}'.format(
+        number=location.prisoner_number.lower(),
+        dob=location.prisoner_dob.strftime('%d/%m/%Y')
+    )
+    hash_object = hashlib.sha256(original.encode())
+    return hash_object.hexdigest()
```
SteveMarshall: I wonder if we should make this use `_calculate_prisoner_hash` from the model (and maybe make that `PrisonerLocation.generate_hash(number, dob)`)?

mtp_api/apps/prison/models.py
```diff
@@ -18,4 +20,20 @@ class PrisonerLocation(TimeStampedModel):
...
+    # SHA256 of prisoner_number + prisoner_dob to that it can
+    # be used for db joins
+    prisoner_hash = models.CharField(max_length=250)
+
     prison = models.ForeignKey(Prison)
+
+    def _calculate_prisoner_hash(self):
+        original = '{number}_{dob}'.format(
+            number=self.prisoner_number.lower(),
+            dob=self.prisoner_dob.strftime('%m/%d/%Y')
```
SteveMarshall: I think this should be `%d/%m/%Y`? That's the format that's used above…

marcofucci:
Closing this because we will try with a multicolumn index on `(prisoner_id, prisoner_dob)` first and see how efficient it is.

PR #20, opened 2015-08-11
Add prisoner number/dob index on PrisonerLocation

marcofucci:
This adds a multi column index on <prisoner_number> and <prisoner_dob> to speed up table joins.


marcofucci:
@SteveMarshall @ibrechin don't even know if it's a good idea to do this now or it's too early but I might forget it so, here it is :)

SteveMarshall:
LGTM! :+1:

ibrechin:
:+1:

PR #21, opened 2015-08-12
Add sender fields to the Transaction object

marcofucci:
This adds sender sort code, account nr and name as new fields of the Transaction model.
I've also got rid of the test_transactions fixture as we don't want to keep updating it every time the models change.


marcofucci:
@SteveMarshall @ibrechin I guess this makes sense?

mtp_api/apps/transaction/models.py
```diff
@@ -14,6 +14,7 @@
 
 
 class Transaction(TimeStampedModel):
+    # TODO, not sure we need this, we might decide to delete it
     upload_counter = models.PositiveIntegerField()
```
SteveMarshall: I wonder if we should replace this with a list of hashes of uploaded transaction files so far, or something, so we can avoid double uploads? Totally separate issue, though…

mtp_api/apps/transaction/models.py
```diff
@@ -22,16 +23,20 @@ class Transaction(TimeStampedModel):
...
-        blank=True, max_length=250
-    )
-    sender_customer_reference = models.CharField(
-        blank=True, max_length=250
-    )
+
+    # cannot be empty otherwise we can't send the money back
+    sender_sort_code = models.CharField(max_length=50)
+    sender_account_nr = models.CharField(max_length=50)
+    sender_name = models.CharField(max_length=100)
```
SteveMarshall: Is the `sender_name` length from the BAI2 spec? Might be worth considering a longer field length, or making a text field because some people might have crazy long names…

SteveMarshall:
We probably need to think about storing the roll number for building societies, but maybe that's a job for another PR?

mtp_api/apps/transaction/models.py
marcofucci: agree I'll set to 250

mtp_api/apps/transaction/models.py
marcofucci: yeah let's keep it as it is and thing about it later on

marcofucci:
I could add an optional `sender_roll_number` to model and API? would it be enough?

SteveMarshall:
@marcofucci Sounds good to me; wanna add `sender_roll_number` in this PR?

marcofucci:
I've decided to just leave the `sender_roll_number` out, we just don't know enough yet.
@SteveMarshall Can you take a final look?

ibrechin:
@marcofucci @SteveMarshall this looks good to me,  is it ready for merge?

marcofucci:
@SteveMarshall ready for merge

PR #22, opened 2015-08-17
Create transactions endpoint

marcofucci:
This adds a new bank-admin namespaced transactions endpoint with the ability
of creating transactions if you have the right permissions.
This also adds a new BankAdmin group with just POST permissions on
bank-admin/transactions/.

### DIFF
--- a/mtp_api/apps/mtp_auth/fixtures/initial_groups.json
+++ b/mtp_api/apps/mtp_auth/fixtures/initial_groups.json
@@ -14,7 +14,9 @@
   "pk": 2,
   "fields": {
     "name": "BankAdmin",
-    "permissions": []
+    "permissions": [
+      ["add_transaction", "transaction", "transaction"]
+    ]
   }
 },
 {
--- a/mtp_api/apps/transaction/serializers.py
+++ b/mtp_api/apps/transaction/serializers.py
@@ -1,5 +1,9 @@
+from django.db import transaction
+from django.db.models import Max
+
 from rest_framework import serializers
 
+from .signals import transaction_created
 from .models import Transaction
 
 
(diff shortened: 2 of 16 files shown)


marcofucci:
@SteveMarshall @ibrechin can you take a look?

SteveMarshall:
Would've been nice to see this as a few separate commits (one to split out the transaction log tests, another for the router changes, and so on), but other than that, LGTM! @ibrechin?

mtp_api/apps/mtp_auth/fixtures/initial_groups.json
```diff
@@ -14,7 +14,9 @@
   "pk": 2,
   "fields": {
     "name": "BankAdmin",
-    "permissions": []
+    "permissions": [
+      ["add_transaction", "transaction", "transaction"]
```
SteveMarshall: Isn't one of these fields a description?

mtp_api/apps/transaction/serializers.py
```diff
@@ -31,3 +35,48 @@ class Meta:
...
         )
+
+
+class CreateTransactionListSerializer(serializers.ListSerializer):
+
+    @transaction.atomic
+    def create(self, validated_data):
+        upload_counter = Transaction.objects.all().aggregate(
+            Max('upload_counter')
+        )['upload_counter__max'] or 1
```
SteveMarshall: Do we risk a race condition here? Is it worth ditching the counter in favour of a UUID or something?

mtp_api/apps/mtp_auth/fixtures/initial_groups.json
marcofucci: nope

mtp_api/apps/transaction/serializers.py
marcofucci: yes possible race condition but unlikely to happen IMHO so not worth spending time on it unless you can think of any real case when this could happen.

Also, still not sure we really need an `upload_counter`, probably we don't. We can have a chat this or next week maybe and decide?

ibrechin:
I've got no objections beyond those questions which we agreed to defer.

@marcofucci @SteveMarshall happy to merge if ready.

PR #23, opened 2015-08-18
Per application permissions

marcofucci:
This adds per-application permissions to the transaction and prisoner locations views so that they can't be accessed by users who logged in using a different oauth application.

- Oauth application ids are now constants so that we can reference them during our tests and when live without hardcoding them every time.
- Replaced the `force_authenticate` call with a token-based one during tests so that we can reliably test the authentication/permissions bit

### DIFF
--- a/mtp_api/apps/mtp_auth/tests/utils.py
+++ b/mtp_api/apps/mtp_auth/tests/utils.py
@@ -14,11 +14,12 @@ class AuthTestCaseMixin(object):
         'PrisonClerk': CASHBOOK_OAUTH_CLIENT_ID
     }
 
-    def _get_http_authorization_token_for_user(self, user):
-        group = user.groups.first()
-        if not group:
-            return None
-        client_id = self.APPLICATION_ID_MAP.get(group.name)
+    def _get_http_authorization_token_for_user(self, user, client_id=None):
+        if not client_id:
+            group = user.groups.first()
+            if not group:
+                return None
+            client_id = self.APPLICATION_ID_MAP.get(group.name)
         application = Application.objects.get(client_id=client_id)
 
         token = get_random_string()
@@ -30,6 +31,8 @@ class AuthTestCaseMixin(object):
         )
         return token
 
-    def get_http_authorization_for_user(self, user):
-        token = self._get_http_authorization_token_for_user(user)
+    def get_http_authorization_for_user(self, user, client_id=None):
+        token = self._get_http_authorization_token_for_user(
+            user, client_id=client_id
+        )
         return "Bearer {0}".format(token)
--- a/mtp_api/apps/mtp_auth/tests/utils.py
+++ b/mtp_api/apps/mtp_auth/tests/utils.py
@@ -1,0 +1,35 @@
+import datetime
+from django.utils import timezone
+from django.utils.crypto import get_random_string
+
+from oauth2_provider.models import Application, AccessToken
+from mtp_auth.constants import CASHBOOK_OAUTH_CLIENT_ID, \
+    BANK_ADMIN_OAUTH_CLIENT_ID, PRISONER_LOCATION_OAUTH_CLIENT_ID
+
+
+class AuthTestCaseMixin(object):
+    APPLICATION_ID_MAP = {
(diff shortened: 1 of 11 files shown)


marcofucci:
@SteveMarshall @ibrechin can you take a look?
Might be worth taking a look at the commits first to understand what I had to do.

marcofucci:
@SteveMarshall @ibrechin ?

mtp_api/apps/transaction/permissions.py
```diff
@@ -12,7 +14,12 @@ def has_permission(self, request, view):
 class IsOwnPrison(BasePermission):
 
     def has_permission(self, request, view):
-        return request.user.prisonusermapping.prisons.filter(pk=view.kwargs.get('prison_id')).exists()
+        try:
+            prison_user_mapping = request.user.prisonusermapping
+        except PrisonUserMapping.DoesNotExist:
+            return False
```
SteveMarshall: Feels a bit clunky to use a try/except here; is there a way we could check instead of waiting for an exception?

mtp_api/apps/mtp_auth/tests/utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+class AuthTestCaseMixin(object):
+    APPLICATION_ID_MAP = {
+        'PrisonerLocationAdmin': PRISONER_LOCATION_OAUTH_CLIENT_ID,
+        'BankAdmin': BANK_ADMIN_OAUTH_CLIENT_ID,
+        'PrisonClerk': CASHBOOK_OAUTH_CLIENT_ID
+    }
+
+    def _get_http_authorization_token_for_user(self, user):
+        group = user.groups.first()
```
SteveMarshall: This line makes me wonder: is it worth adding some tests for users in more than one group, to ensure they can't access an app's endpoints when using a different one… Or is that covered enough by `ClientIDPermissionsTestCase. test_no_permissions_with_mismatching_clients`?

mtp_api/apps/transaction/permissions.py
marcofucci: The only other way is to make an extra query which doesn't seem great.
This actually happen all the time in python, try/except exceptions is not as bad as it seems in other languages.

mtp_api/apps/prison/tests/test_views.py
```diff
@@ -26,12 +28,35 @@ def setUp(self):
...
 
-    def test_fails_without_permissions(self):
-        unauthorised_user = self.prison_clerks[0]
+    def test_fails_without_application_permissions(self):
+        """
+        Tests that if the user logs in via a different application,
+        they won't be able to access the API.
+        """
+        for unauthorised_user in [
+            self.prison_clerks[0], self.bank_admins[0]
```
SteveMarshall: What about testing valid users of this tool? If they're a valid user but accessing from another app, we don't want that to work…

mtp_api/apps/transaction/permissions.py
SteveMarshall: Fair enough! :+1:

mtp_api/apps/prison/tests/test_views.py
marcofucci: Good point, it might be worth adding. Not sure if I can make it elegant enough but I'll do my best

mtp_api/apps/mtp_auth/tests/utils.py
marcofucci: Would pretty much get covered by https://github.com/ministryofjustice/money-to-prisoners-api/pull/23#discussion_r37415654 I guess?

ibrechin:
Other than perhaps adding the suggested tests, this looks good to me.

mtp_api/apps/mtp_auth/tests/utils.py
SteveMarshall: Yeah, think so?

marcofucci:
@SteveMarshall cool?

SteveMarshall:
@marcofucci :snowflake:

PR #25, opened 2015-08-20
Create Cashbook-transactions-API-design.md

marcofucci:


marcofucci:
Moved from gist to PR as gist don't work well (they don't support notifications for example, thanks GitHub!)

marcofucci:
@SteveMarshall said:

For the most part, I think this is great. A few small observations/thoughts:

1. The filters on `/cashbook/transactions/` are good, but I think `status` is missing a `refunded` value (and maybe `refund_pending`, too?). I'd also argue that `available` is the only status for which filtering by user isn't applicable: `pending`, `credited`, and `refunded` should all have assigned users, so filtering by user should return all of those. It would also be nice to allow multiple user IDs to be selected, long term.
2. We should probably standardise the language we use for locking/taking transactions. I can see three different terms being used here: `take`, `pending`, and `locked`. My preference would be `lock`/`locked` throughout, but I'm open to other terms…
3. The explanation of `count` on `…/take/` isn't entirely clear to me. Do we have a hard limit of 20, or does specifying `count=50` mean that I get a total of 50 locked to me? If the latter, what happens if I have 50 locked and then try to lock with `count=20`? Do we release the excess?
4.  Do we need `prison` on `…/take/` to be mandatory? My instinct is not: if someone only has permission to process one prison, that's all they'll get. If they have more than one prison, chances are they don't care which prison they're crediting to when crediting (and, if they *do*, that's something we can handle in the cashbook).
5. Leave out the `for_user` stuff on `…/take/`: we don't need it yet, as you say.
6. Should the 400 returns in `/transaction/ PATCH` and in `…/release/` be `409 Conflict`, maybe?

marcofucci:
Re. point 2, so we decided that the action should be `lock` and the status should be `locked`.

marcofucci:
@SteveMarshall @ibrechin So the status `available` becomes `unlocked`? Does it make sense?

SteveMarshall:
@marcofucci I think `available` stays as it is: `unlocked` only makes sense to me if it was `locked` first?

marcofucci:
Let's see point 4.

> Do we need prison on …/take/ to be mandatory? My instinct is not: if someone only has permission to process one prison, that's all they'll get. If they have more than one prison, chances are they don't care which prison they're crediting to when crediting (and, if they do, that's something we can handle in the cashbook).

I'm cool with `if someone only has permission to process one prison, that's all they'll get` but not sure about what happens with multiple prisons.

At the moment, the cashbook frontend does not show the prison anywhere and I'm worried that if somebody adds a prison to a user and the cashbook frontend app doesn't show that data, they would get confused. Or we just don't care?

What do you think @SteveMarshall @ibrechin ?

SteveMarshall:
I think we don't care, actually: if I remember it correctly, NOMIS will happily let the user credit prisoners from all over the place in one batch. The only reason we have to break it up by prison is because of NOMS finance rules, and our tools will handle that stuff for them…

marcofucci:
If we don't need `prison` on `/lock` at the moment, let's get rid of it, no?

SteveMarshall:
:+1:

marcofucci:
@ibrechin are you OK with removing the prison param from the /lock action?

ibrechin:
yes

marcofucci:
point 3:

> The explanation of count on …/take/ isn't entirely clear to me. Do we have a hard limit of 20, or does specifying count=50 mean that I get a total of 50 locked to me? If the latter, what happens if I have 50 locked and then try to lock with count=20? Do we release the excess?

As it seems confusing and unnecessary maybe we could just remove it and:
- don't do anything if they have already locked some (even if just one)
- lock 20 if they haven't been locking any

Not sure I like it but open to suggestions.

SteveMarshall:
I reckon kill the `count` parameter but always try to return a batch of 20.

ibrechin:
Seems fair enough to just always do what you specified as the default behaviour.

marcofucci:
Now let's see point 6

> Should the 400 returns in /transaction/ PATCH and in …/release/ be 409 Conflict, maybe?

The w3 specs say about `409`:

> The request could not be completed due to a conflict with the current state of the resource. This code is only allowed in situations where it is expected that the user might be able to resolve the conflict and resubmit the request.

So my interpretation is that `409` should be used in cases where the user can change the resource and submit the request again.

Re. `/unlock/`, there's nothing that the user can do to resolve the conflict as the transactions are credited. Therefore, I think that `400` works better than `409`.

Re. `PATCH` on `/transactions/` (to mark them as credited), there's nothing that the user can do either as the API don't allow locking specific resources so I think that `400` works better than `409` in this case as well.

What do you guys think?

SteveMarshall:
I think it depends how you define the user? The “user” of the API is the cashbook (or, rather, developers consuming the API), to my mind, so `409` is appropriate.

If we're considering the user of the consuming app the user of the API, though, I agree that `400` is the code to use. In future, (when we no longer have our own auth provider hosted in the API) nothing but apps we build should talk to the API, so I'd argue the humans using the cashbook (or whatever) aren't *really* the “user”.

marcofucci:
@SteveMarshall even if the user of the API is the cashbook, it seems like `409` is not appropriate. How would they be able to resolve those conflicts?

ibrechin:
Whatever you decided in terms of status code, I feel like both the cases listed in /transactions/ PATCH should return the same. Whether it's locked by another user or not locked at all, the basic reason it doesn't work is, in both cases, because the transaction is not locked to the requesting user. Could also argue that having different responses is leaking information, although that is not a major concern considering it's a private API.

marcofucci:
@ibrechin might be easier if we think that way maybe. So the question is, what should it return if at least one of the transactions have not been locked by the user first?

ibrechin:
Could make the argument for 409 that as it is triggered on an at-least-one basis, the user could resolve the conflict by removing the transactions it is not allowed to credit from the list of ids.

403 would also be perfectly fine. I don't think 400 is suitable, though, as the request is basically valid, just not able to be fulfilled due to current server state.

marcofucci:
Ok so to recap:

## /transactions/ -- PATCH

- returns 409 if at least one of the transactions have not been locked by the user.

## /transactions/actions/unlock/ -- POST

- returns 409 if at least one of the transactions have not been locked by the user.


@SteveMarshall @ibrechin OK?

ibrechin:
@marcofucci I don't think we can use the same logic for /unlock, as it does not follow the same rules - in the current version of the document, it has requirements that the user must own the prison, and transactions cannot have been credited.

marcofucci:
@ibrechin I think they are actually the same. you can PATCH on `/transactions/` passing ids of transactions that belong to different prison as well.
Looks like exactly the same logic.

ibrechin:
@marcofucci wasn't there a requirement that users be able to unlock transactions that have been locked by someone else? Use case being someone locks transactions, forgets to unlock, and then goes home for the day.

marcofucci:
@ibrechin yes but that is all filter by prisons that you can manage meaning user managing prisons A and B can unlock any transactions for prison A or B

marcofucci:
Ok so to recap:

## /transactions/ -- PATCH

- returns 403 if at least one of the transactions belongs to a prison not managed by the logged-in user
- returns 409 if at least one of the transactions cannot be credited

## /transactions/actions/unlock/ -- POST

- returns 403 if at least one of the transactions belongs to a prison not managed by the logged-in user
- returns 409 if at least one of the transactions cannot be unlocked

@SteveMarshall @ibrechin better?

ibrechin:
Looks legit.

marcofucci:
Let's see point 1. now

> The filters on `/cashbook/transactions/` are good, but I think `status` is missing a `refunded` value (and maybe `refund_pending`, too?). I'd also argue that `available` is the only status for which filtering by user isn't applicable: `locked`, `credited`, and `refunded` should all have assigned users, so filtering by user should return all of those. It would also be nice to allow multiple user IDs to be selected, long term.

* I'm OK with adding `refunded` and `refund_pending` to the doc and mark those as `not implemented yet`.
* I might be OK with multiple user IDs but I'd say, if it's too complicated to implement, we won't implement it as I don't think we'll ever use it.

In general, the endpoint seems a bit messy with all those filters mixed up a bit but if you guys are OK with it, I am too.

@SteveMarshall @ibrechin ?

SteveMarshall:
Yeah, I think multiple ID support is definitely a far-future-when-we-actually-need-it thing. :+1: on `not implemented yet` for `refunded` and `refund_pending`.

ibrechin:
I'm fine with the suggested implementation.

On the user/status thing, I think it's simpler just to say that it returns all transactions attached to that user, without considering status. There may be certain statuses which mean that the transaction is not attached to *any* user, but this isn't relevant from the perspective of the API - all we care is that it isn't attached to the particular user we're querying for.

marcofucci:
Cool, it looks a lot better now.
The only thing I don't really like too much is `/transactions/` without `status` param but with `user`.
In theory, it should return all transactions excluding the ones that are 'attached' to other users.
Might be a bit hard to implement in an elegant way but definitely doable.

Do you guys agree it should return that?

SteveMarshall:
That feels correct to me, yeah.

SteveMarshall:
And, yeah, this is looking much, much cleaner. Good work! :stars:

ibrechin:
What is the actual use case for when someone would query transactions filtered by user?

ibrechin:
Unless this doesn't satisfy the use case, I think it is most consistent if filtering by user doesn't return unattached transactions. If you've provided a user filter, then I assume you're interested in what that user is involved in, not in the mass of available transactions.

marcofucci:
@ibrechin when a user logs in, they want to see just the transactions locked by them so they'll call

`/cashbook/transactions/?user=<me>&status=locked`

They also want to know how many transactions are still available so they'll call:

`/cashbook/transactions/?status=available`

There's also another panel to show transactions what have been locked by anyone so they'll call:

`/cashbook/transactions/?status=locked`

In the near future, a user should also be able to get an history of the transactions they've credited so they'll call:

`/cashbook/transactions/?user=<me>&status=credited`.

I'm not sure we need anything else at the moment.

ibrechin:
Thanks, that's very helpful. In that case I'd say go with whatever you want/is easiest, and if ever someone does come up with a use case that involves querying just on user we can change the behaviour if it doesn't match what they want.

SteveMarshall:
:+1:

marcofucci:
Cool, I might leave this as it is now, start implementing it on Monday and see if I spot anything weird before merging it so that we keep the discussion here.

Thanks guys!

SteveMarshall:
We don't need these design docs anymore. Closing.

PR #26, opened 2015-08-24
Bank admin API design doc

ibrechin:
Suggested API for the bank admin endpoint. Attempted to adhere to the conventions we established in the cashbook API design.

@marcofucci @SteveMarshall comments?

### DIFF
--- a/Bank-admin-API-design.md
+++ b/Bank-admin-API-design.md
@@ -37,7 +37,7 @@ List of transaction elements, which must consist of the following two values:
 - *id*: id of the transaction to be changed
 - *refunded*: `True` if the transaction has been refunded, `False` otherwise
 
-## /bank_admin/transactions/ - PUT
+## /bank_admin/transactions/ - POST
 
 Create new transactions
 
--- a/Bank-admin-API-design.md
+++ b/Bank-admin-API-design.md
@@ -8,8 +8,8 @@ Returns a list of all recent transactions
 
 **status**
 
-Filters by the status of the transactions, accepts multiple values which will be 
-OR'd.
+Filters by the status of the transactions, accepts a comma separated list
+of values which will be OR'd.
 
 Possible values:
 - *available*: returns list of transactions that have not been locked by anyone
@@ -20,7 +20,7 @@ Possible values:
 
 For ADI file generation, this will likely be:
 
-?status=credited&status=refunded
+?status=credited,refunded
 
 For refund CSV generation, this will likely be:
 
--- a/Bank-admin-API-design.md
+++ b/Bank-admin-API-design.md
@@ -1,0 +1,55 @@
+# Bank admin API
+
+## /bank_admin/transactions/ - GET
+
+Returns a list of all recent transactions
+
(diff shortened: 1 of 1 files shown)


ibrechin:
1. Sure
2. I prefer hypens in URLs, but I changed to underscore for compatibility with slumber (you can't do client.bank-admin()
3. My understanding was that it would be the same people doing both these tasks, and thus a bank-admin user would have permissions for both, on which basis it doesn't matter if all information is returned in both cases. If this is not the case and we don't want the people doing the ADI file processing to have the bank details then this can be changed, otherwise I think it's fine for the client to just take what it needs.

marcofucci:
2. right, let's keep it as it is
 3. I think there will be different people. We can have serialisers depend on permissions and return full or partial data as Steve suggested or have separate endpoints, not sure

Bank-admin-API-design.md
```diff
@@ -0,0 +1,55 @@
...
+Possible values:
+- *available*: returns list of transactions that have not been locked by anyone
+- *locked*: returns list of locked transactions
+- *credited*: returns list of transactions credited
+- *refunded* (*not implemented yet*): returns list of transactions refunded
+- *refund_pending* (*not implemented yet*): returns list of transactions in pending refund
+
+For ADI file generation, this will likely be:
+
+?status=credited&status=refunded
```
SteveMarshall: Wait, are we saying `&` is treated at `OR`? That seems counter-intuitive? I'd suggest we treat `&` as `AND`, and use comma delimiting for `OR` (i.e. `?status=credited,refunded`).

SteveMarshall:
1. :+1: for the `POST` change.
2. I'm not a big fan of the underscores but, yeah, if that makes slumber integration bad, then let's leave it. (Am I right in thinking, though, that we could do `client('bank-admin')` if we really, really wanted underscores?)
3. As for ADI/AccessPay permissions: I *think* the two are generated by the same people at the moment, and those people then pass the AccessPay file off to another team to process it, but this is something we're waiting for SSCL to confirm. Let's work on the assumption it's different people, because that's what it *should* be.

Bank-admin-API-design.md
ibrechin: The & is not boolean, it's just the URL param separator. Multiple values of the same param is the easiest way (imo) of submitting a list of values via GET, and avoids having to parse a string.

SteveMarshall: Yeah, I know `&` in a URL has nothing to do with boolean `AND`. My point is more that adding extra filters (in our case, URL params) would normally imply that you're restricting the resultset further, rather than expanding it.

ibrechin: I just don't like trying to render data structures in text fields. In this case it is unlikely that any problems will arise as the values of status are limited, so if the & bothers you I will change it.

ibrechin:
I've been thinking about 3.

It seems like the rationale for it is the same as that we

SteveMarshall:
We don't need these design docs anymore. Closing.

PR #27, opened 2015-08-24
Transactions api split

marcofucci:
So that it's clearer and different people can work on different files at the same time


marcofucci:
@SteveMarshall @ibrechin what do you think?
This would allow @ibrechin and I to work on different parts at the same time.
It is also cleaner I guess

SteveMarshall:
Looks much cleaner to me! @ibrechin?

ibrechin:
I like.

PR #28, opened 2015-08-27
Cashbook transactions v2

marcofucci:
New version of the cashbook transactions API we decided to re-design using: 

https://github.com/ministryofjustice/money-to-prisoners-api/blob/Cashbook-transactions-API-design/Cashbook-transactions-API-design.md


marcofucci:
@SteveMarshall @ibrechin code review? Please don't merge it just yet as the cashbook app is not ready

mtp_api/apps/transaction/api/cashbook/views.py
```diff
@@ -50,117 +53,131 @@ def get_prison_set_for_user(self):
...
         transaction_ids = request.data.get('transaction_ids', [])
         with transaction.atomic():
-            to_update = self.get_queryset().pending().filter(pk__in=transaction_ids).select_for_update()
+            to_update = self.get_queryset().locked().filter(pk__in=transaction_ids).select_for_update()
             if len(to_update) != len(transaction_ids):
                 return Response(
-                    data={'transaction_ids': ['Some transactions could not be released.']},
-                    status=status.HTTP_400_BAD_REQUEST
+                    data={'transaction_ids': ['Some transactions cannot be unlocked.']},
+                    status=drf_status.HTTP_409_CONFLICT
```
SteveMarshall: Can we list the transaction IDs that couldn't be unlocked? Maybe have the current message as `error`, then `failed_transaction_ids` array?

mtp_api/apps/transaction/tests/test_cashbook_views.py
```diff
@@ -36,800 +34,584 @@ def _get_authorised_user(self):
...
+    def test_WITH_prison_AND_user_DEFAULT_status(self):
+        self._test_response_with_filters(filters={
+            'prison': self.prisons[0].pk,
+            'user': self.prison_clerks[1].pk
+        })
+
+    def test_WITH_user_DEFAULT_status_prison(self):
+        self._test_response_with_filters(filters={
+            'user': self.prison_clerks[1].pk
+        })
```
SteveMarshall: These tests feel like prime candidates for the sort of refactor we've been discussing in https://github.com/ministryofjustice/money-to-prisoners-bank-admin/pull/3#issuecomment-135744749. Maybe test cases of `TransactionListWithDefaultPrisonAndUser`, `TransactionListWithDefaultPrison`, and `TransactionListWithDefaultUser`… or something. Waddya think?

mtp_api/apps/transaction/api/cashbook/views.py
```diff
@@ -50,117 +53,131 @@ def get_prison_set_for_user(self):
...
-    queryset = Transaction.objects.all()
+class GetTransactions(TransactionViewMixin, generics.ListAPIView):
     serializer_class = TransactionSerializer
-    patch_serializer_class = CreditedOnlyTransactionSerializer
-    filter_backends = (
-        filters.DjangoFilterBackend,
-        filters.OrderingFilter,
+    filter_backends = (filters.DjangoFilterBackend,)
+    filter_class = TransactionListFilter
+    action = 'list'
```
SteveMarshall: This declarative specification of the action for permissions is neat :+1:

mtp_api/apps/transaction/api/cashbook/views.py
marcofucci: could do but they we'd been leaking data. I could try calling the endpoint with id from 1 to n and get back the invalid (and therefore the invalid) ones in one go without too much noise.

mtp_api/apps/transaction/tests/test_cashbook_views.py
marcofucci: I actually like it like this as you can see the differences between them easily and they are basically the same + function param which shows you exactly the inputs we are using for the tests.

Splitting them in different classes with just one function in them would be messy for me.
Don't you think?

mtp_api/apps/transaction/api/cashbook/views.py
SteveMarshall: Could we do something like getting the difference of the two sets of ids? Something like:

```python
ids_to_update = [t.id for t in to_update]
conflict_ids = set(transaction_ids) - set(ids_to_update)
```

marcofucci: Does it mean that you don't agree with leaking data being a problem?

SteveMarshall: We're not saying anything about the transactions, just that the calling user couldn't unlock them, right? I don't think that info's *too* sensitive?

marcofucci: yeah being too paranoid I guess, will change this :+1:

mtp_api/apps/transaction/tests/test_cashbook_views.py
SteveMarshall: Sorry, I wasn't entirely clear: I'm not saying one class per function here. Something a bit like this (with details omitted for clarity:

```python
class TransactionListWithDefaultPrisonAndUserTestCase(…):
    def test_no_filter_returns_all_transactions(…):
        # …
    def test_filter_by_available_status_returns_only_available_transactions(…):
        # …
    def test_filter_by_credited_status_returns_only_credited_transactions(…):
        # …
```

… and so on.

That way, much of the initial setup logic in `_test_response_with_filters` can be moved into class-level `setUp` methods (maybe with a base class for all the tests or something), and each test method can directly make the request and test the responses.

marcofucci:
@SteveMarshall @ibrechin made the suggested changes, does it look OK now?

mtp_api/apps/transaction/api/cashbook/views.py
```diff
@@ -50,117 +53,151 @@ def get_prison_set_for_user(self):
...
+            ids_to_update = [t.id for t in to_update]
+            conflict_ids = set(transaction_ids) - set(ids_to_update)
 
-        return qs
+            if conflict_ids:
+                return Response(
+                    data={
+                        'error': [
+                            'The following transactions cannot be credited: {ids}'.format(
+                                ids=', '.join(sorted([str(t_id) for t_id in conflict_ids]))
```
ibrechin: I think it would be more machine-readable if the conflicted IDs were returned in their own JSON attribute rather than as part of the error message string.

SteveMarshall: :+1:

marcofucci: @ibrechin what format did you have in mind?

Something like this?

    "error": [
        "1": "this transaction cannot be credited",
        "2": "this transaction cannot be credited",
        "3": "this transaction cannot be credited"
    ]

?

SteveMarshall: @marcofucci I'd tend toward structuring it [like Gocardless's errors](https://github.com/gocardless/http-api-design#nested-errors)

marcofucci: @SteveMarshall well, I guess @ibrechin meant this https://github.com/ministryofjustice/money-to-prisoners-api/pull/29/files#diff-f953809aae8df9028561468f64e35b88R50

ibrechin: @marcofucci I was actually thinking

    "error": ["Some transactions could not be credited"]
    "ids": [1,2,3]

...although on reflection this prevents you from having multiple errors applying to different transactions, if that's a thing we want. Possibly:

    "error": [
        {
        "msg": "Some transactions could not be credited"
        "ids": [1,2,3]
        }
    ]

marcofucci: "error": [
        {
        "msg": "Some transactions could not be credited"
        "ids": [1,2,3]
        }
    ]

looks fine to me, what do you think, @SteveMarshall ?

SteveMarshall: Make the key `errors`, and I'm sold :+1:

ibrechin:
All fine with me.

PR #29, opened 2015-08-27
Bank admin transactions

ibrechin:
Implementation of the bank admin API design


ibrechin:
@marcofucci @SteveMarshall up for review.

There will undoubtedly be conflicts with https://github.com/ministryofjustice/money-to-prisoners-api/pull/28 so I will probably rebase this after that one is merged.

mtp_api/apps/transaction/api/bank_admin/serializers.py
```diff
@@ -50,3 +51,75 @@ class Meta:
...
+            pk__in=to_refund,
+            **Transaction.STATUS_LOOKUP['refund_pending']).select_for_update()
+        if len(update_set) != len(to_refund):
+            raise Transaction.DoesNotExist(
+                list(set(to_refund) - {t.id for t in update_set})
+            )
+
+        update_set.update(refunded=True)
+        for t_id in to_refund:
+            transaction = Transaction()
```
marcofucci: I think this might be dangerous as the signal handler wouldn't know that the transaction object only has the ID set.

ibrechin: Fair.

mtp_api/apps/transaction/tests/test_bank_admin_views.py
```diff
@@ -82,3 +85,294 @@ def test_create_list(self):
...
+        self.assertEqual(current_count, Transaction.objects.count())
+
+    def test_create_list_null_account_number_bad_request(self):
+        self._test_create_list_incomplete_data('sender_account_number')
+
+    def test_create_list_null_sort_code_bad_request(self):
+        self._test_create_list_incomplete_data('sender_sort_code')
+
+    def test_create_list_null_amount(self):
+        self._test_create_list_incomplete_data('amount')
```
SteveMarshall: Can you rename these to `test_create_list_with_[…]_fails`, and pull the body of the test back into the individual test cases? It'd be much clearer to have them that way than with the assertions farmed off to the other function. You could have a `_get_data_list_with_null_field` method that accepts the field to null as an argument, though: I think that'd be pretty clear.

mtp_api/apps/transaction/tests/test_bank_admin_views.py
```diff
@@ -82,3 +85,294 @@ def test_create_list(self):
...
+                    Transaction.objects.get(id=data['id']).refunded
+                )
+
+    def test_patch_credited_creates_conflict(self):
+        self._test_patch_invalid_refunds_creates_conflict('credited')
+
+    def test_patch_refunded_creates_conflict(self):
+        self._test_patch_invalid_refunds_creates_conflict('refunded')
+
+    def test_patch_cannot_update_disallowed_fields(self):
```
SteveMarshall: These test names are great; be good to see the body of the tests in here, though, as I said on `_test_create_list_incomplete_data`.

mtp_api/apps/transaction/tests/utils.py
```diff
@@ -60,7 +60,7 @@ def choose_user(self, prison):
     now = timezone.now().replace(microsecond=0)
 
     for upload_counter in range(1, uploads+1):
-        for transaction_counter in range(1, transaction_batch+1):
+        for i, ransaction_counter in enumerate(range(1, transaction_batch+1)):
```
SteveMarshall: I think `ransaction_counter` is missing a leading `t`?

SteveMarshall: That said, is it used? Could it just be `_`?

ibrechin: Actually the entire change is pointless, I clearly wasn't paying attention. Why did I enumerate a range?

mtp_api/apps/transaction/urls.py
```diff
@@ -11,9 +11,10 @@
 cashbook_transaction_router.register(r'transactions', cashbook_views.TransactionView)
 
 admin_transaction_router = routers.DefaultRouter()
-admin_transaction_router.register(r'transactions', bank_admin_views.TransactionView)
+admin_transaction_router.register(r'transactions', bank_admin_views.TransactionView,
+                                  base_name='transaction')
 
 urlpatterns = patterns('',
-    url(r'^bank-admin/', include(admin_transaction_router.urls, namespace='bank-admin')),
+    url(r'^bank_admin/', include(admin_transaction_router.urls, namespace='bank_admin')),
```
SteveMarshall: Damn you, Slumber! :angry: (This is fine, but our poor URLs suffer.)

mtp_api/apps/transaction/tests/utils.py
SteveMarshall: A good question! You can, of course, act like it never happened by [doing a fixup commit](https://robots.thoughtbot.com/autosquashing-git-commits) against 81e6326.

SteveMarshall: (Which, by the way, you should push so we can review the changes, then rebase against master once everyone's happy.)

SteveMarshall:
This PR has a really good, clear commit history, by the way :star:

mtp_api/apps/transaction/tests/test_bank_admin_views.py
ibrechin: Renaming is fine.

What is your rationale for increasing code duplication in this case? I can see the reasoning behind e.g. not making a test just a list of helper methods, because then the test logic is scattered in multiple locations, but this is really just a single test that happens to be being run with different inputs. If you want to know what happens in the test case you only have one place to look.

SteveMarshall: I'm not sure I agree that it's a single test with different inputs: it certainly appears that way now, but they're three distinct fields, and could all have different behaviours in future. The current tests imply that there's some sort of intentional link between the result of the three tests, when they actually just happen to have the same response to invalid input across three distinct fields.

If we were testing multiple invalid inputs to a single field, I'd probably agree that it's the same test (though I'd probably suggest using a [dataprovider-type pattern](https://pypi.python.org/pypi/unittest-data-provider/1.0.0) would be a clearer way to represent that).

I'd suggest that having a helper to null a given data list field is a more appropriate abstraction that reduces the duplication enough that doesn't imply intent in the result of the call to the API. The methods could also have the `url` and `user` objects inlined into the `self.client.post()` and be slightly less duplicate-y, I think?

ibrechin: I don't think they "happen" to have the same result - the test is based on recognising that a required field is absent, and those are all required fields.

I don't really like the linked dataprovider thing. I think adding syntactic sugar for calling a function multiple times with different values is overkill, and it also prevents you from naming tests according to the rules that you have been espousing i.e. specifying inputs in the method name.

How about if I move just the assertions out of the base function and into the callers? That way if the expected per-field behaviour changes it's easy to alter, but it doesn't duplicate the whole "setting a field to null and making a create request with it" thing which is unlikely to differ?

marcofucci: I don't like the dataprovider thing either. I think that this is really a minor thing so could easily stay as it is.
If Steve prefers though, moving the assertions out of the base function looks good too.

SteveMarshall: Let's go with that: rename the tests to be more explicit, and pull the assertions into each test. I'm still not convinced that's the right way to do it, but let's go with that.

ibrechin:
@marcofucci @SteveMarshall rebased

SteveMarshall:
@marcofucci Merge if you're happy with this.

PR #30, opened 2015-09-03
Cast ids to int when locking

marcofucci:
The cashbook is sending ids as strings but the API is expecting integers.
I could have changed the cashbook but I've noticed that we are using ints everywhere else so I thought it would be cleaner to cast ids to integers in the API unlock view.


marcofucci:
@ibrechin @SteveMarshall quick look/merge? This basically just casts ids to integers, doing it with serializers makes it slightly cleaner.

SteveMarshall:
Does anything else accept ints? I wonder if we should standardise on one or the other, and update everything to match?

marcofucci:
Every other view casts to int so this is the only place we have to change if I'm not wrong

marcofucci:
@ibrechin rebased and made the change you mentioned

ibrechin:
:+1:

PR #31, opened 2015-09-07
Fail safe if logged user doesn't manage any prisons

marcofucci:
This fixes a problem with users (e.g. bank admins) without PrisonUserMapping model. The code is not catching the case and failing silently.

### DIFF
--- a/mtp_api/apps/mtp_auth/serializers.py
+++ b/mtp_api/apps/mtp_auth/serializers.py
@@ -2,14 +2,17 @@ from django.contrib.auth.models import User
 
 from rest_framework import serializers
 
+from .models import PrisonUserMapping
+
 
 class UserSerializer(serializers.ModelSerializer):
     prisons = serializers.SerializerMethodField()
 
     def get_prisons(self, obj):
-        return [
-            prison.pk for prison in obj.prisonusermapping.prisons.all()
-        ]
+        try:
+            return list(obj.prisonusermapping.prisons.values_list('pk', flat=True))
+        except PrisonUserMapping.DoesNotExist:
+            return []
 
     class Meta:
         model = User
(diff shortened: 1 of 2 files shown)


marcofucci:
@SteveMarshall @ibrechin review?

mtp_api/apps/mtp_auth/serializers.py
```diff
@@ -2,14 +2,17 @@
...
     prisons = serializers.SerializerMethodField()
 
     def get_prisons(self, obj):
-        return [
-            prison.pk for prison in obj.prisonusermapping.prisons.all()
-        ]
+        try:
+            return list(obj.prisonusermapping.prisons.values_list('pk', flat=True))
+        except PrisonUserMapping.DoesNotExist:
+            return []
```
SteveMarshall: Is it worth having an explicit test case for a user with no prisons?

marcofucci: yep done, that why I changed `test_views` ;-)

SteveMarshall: Well, it's implicitly testing this. I was asking if we should have a test that says “does a user with no prisons get an empty array?”. The current test is just silently `pass`ing over those users.

marcofucci: well. the current test checks "does a user gets back the list of prisons they can manage?" which includes empty list or non-empty list.
I presume it's enough, don't you think?

ibrechin:
fyi i'm :+1:

mtp_api/apps/mtp_auth/serializers.py
SteveMarshall: So it does. That wasn't entirely clear because it does them all in one test. It'd be much clearer as multiple tests, I think.

marcofucci: hmm OK if you really think it's necessary, I'll do it

marcofucci:
@SteveMarshall I've split the tests as you suggested, do you want to double-check?

SteveMarshall:
Looks great, thanks! :+1:

PR #32, opened 2015-09-07
Remove Transaction.upload_counter

marcofucci:
This is because it's not needed so it's better to remove it and simplify things.


marcofucci:
@SteveMarshall @ibrechin review?

SteveMarshall:
:+1:

PR #34, opened 2015-09-11
Correct the URL used for refunding transactions

ibrechin:
In the previous configuration, the route for refunding transactions
was being set to /transactions/patch_refunded rather than the
intended /transactions/.


ibrechin:
@SteveMarshall @marcofucci found this bug whilst doing smoke testing (sorry to say I hadn't before). Turns out the default routing doesn't work like I thought it did. Corrected the unit test to attempt to use the same URL reversal for patch as list/create etc.

Please review when you have a moment.

marcofucci:
:+1:

PR #36, opened 2015-09-16
Change prisoner locations endpoint URL for slumber compatibility

ibrechin:
As slumber works by converting the attributes of the client object
that are accessed into the URL, and an attribute cannot contain a
hyphen, we must change the URL to contain an underscore instead.


ibrechin:
@marcofucci same deal as the bank-admin -> bank_admin change

PR #37, opened 2015-09-17
Transaction.prison updated from prisoner location upload

marcofucci:
This changes the logic behind the Transaction.prison field.

When new transactions get uploaded and created, they don't have the the notion of 'assigned prison'. This information is in an extra file called *prisoner location upload* containing a list of (prisoner_number, prisoner_dob, prison).
The actual prison of a transaction is the left outer join between the tables `transaction` and `prisoner location` on `(prisoner_number, prisoner_dob)`.

I tried changing the queries so that the prison would be calculated by executing the above join every time but it wasn't working well so we decided to pre-calculate the prison and store it in the `Transaction.prison` field instead.

As transactions and prisoner location records are created at different times though, it was hard to determine when to update prison field so we decided to update it either when a new transaction gets created AND/OR when an new prisoner location file gets uploaded.

During this process, we ignore the transactions which have been locked, credited or refunded so that we don't change any work in progress.
Finally, if a transaction gets unlocked without being credited or refunded, we update the prison field to make sure that we have the latest version.

I had to change the test data as we need dummy prisoner location records now when creating dummy transactions.

### DIFF
--- a/mtp_api/apps/transaction/tests/utils.py
+++ b/mtp_api/apps/transaction/tests/utils.py
@@ -3,6 +3,7 @@ import random
 
 from django.utils import timezone
 from django.utils.crypto import get_random_string
+from django.contrib.auth.models import User
 
 from prison.models import Prison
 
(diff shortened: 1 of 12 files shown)


marcofucci:
@ibrechin finally finished, ready to be core reviewed

mtp_api/apps/transaction/tests/utils.py
```diff
@@ -73,47 +43,85 @@ def choose_user(self, prison):
...
+            user, status = clerks_per_prison[prison.pk]
+            return (next(user), next(status))
+        return internal_function
+
+    data_list = generate_initial_transactions_data(
+        tot=transaction_batch
+    )
+
+    location_creator = get_prisoner_location_creator()
+    onwer_status_chooser = get_owner_and_status_chooser()
```
ibrechin: typo

ibrechin:
We might want to think about adding some new DB indexes for that query but that's probably best done in post-release tuning.

:+1:

PR #38, opened 2015-09-18
Reset test data

ibrechin:
Added a conditional to the docker_entrypoint.sh such that if the DOCKER_STATE env variable is set to 'reset_db' then on deployment to an environment the relevant DB will be wiped and the test data loaded.


ibrechin:
@marcofucci what do you think? acceptable way of doing it?

PR #40, opened 2015-09-21
Add general_ledger_code field to prison table

ibrechin:
Phoenix uses a different code to identify prisons to NOMIS,
so this code, the general ledger code, must also be recorded
in the database and returned by the API for the purpose
of reconciliation with Phoenix.


ibrechin:
@marcofucci see what you think

PR #42, opened 2015-09-30
Make the Makefile use docker-machine, and improve instructions

maxf:
- update the Makefile to use it
- update the instructions, and add details

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,23 +3,24 @@ The API Server for the Money to Prisoners Project
 
 ## Dependencies
 ### Docker
-To run this project locally you need to have
+To run this project locally on a Mac you need to have
+[Virtualbox](https://www.virtualbox.org/wiki/Downloads)
 [Docker](http://docs.docker.com/installation/mac/) and
 [Docker Compose](https://docs.docker.com/compose/install/) installed.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -42,17 +43,16 @@ $ make build # initialise the server
 $ make # run the server
 ```
 
-You should be able to point your browser at
-[http://localhost:8000](http://localhost:8000)
-if you're using *boot2docker* then it'll be at the IP of the boot2docker virtual machine.
-You can find it by typing `boot2docker ip` in a terminal. Then visit http://**boot2docker ip**:8000/
+You should be able to point your browser at the VM's url
+You can find it by typing `docker-machine ip` in a terminal. Then visit http://**boot2docker ip**:8000/.
```
SteveMarshall: `http://**boot2docker ip**:8000/` at the end there…

SteveMarshall:
Can you rebase so your commits fit the [alphagov git style guide](https://github.com/alphagov/styleguides/blob/master/git.md)? (The first line should complete the sentence “If applied, this commit will…”, so that currently reads “If applied, this commit will boot2docker is now docker-machine”)

PR #44, opened 2015-10-06
Only allow users associated with a client to log in to it

ibrechin:
This means that only users which has been explicitly linked to
a particular application (e.g. bank admin, cashbook) can authenticate
with that client. This means that a prison clerk cannot log in to
bank admin (unless they have been allowed to).

Uses a new table mtp_auth_applicationusermapping to store the mapping
of users to applications which they can log in to.


ibrechin:
@marcofucci for your consideration

marcofucci:
@ibrechin looks good! Ready to be merged?

ibrechin:
@marcofucci do it

PR #45, opened 2015-10-08
Run a migrate task before attempting to flush when resetting db

ibrechin:
The flush command operates on an explicit list of table so if the
tables defined in the application is different to those present
in the db it will fail. For this reason a migrate must be run
before the flush command to ensure that the db data model matches
that of the application.


ibrechin:
@SteveMarshall @marcofucci this should fix the current problem with the API deployment

PR #47, opened 2015-10-14
Add reconciliation tracking to transaction model

ibrechin:
This consists of a new field, reconciled, which is orthogonal to
the status of the transaction. It is now possible to filter by
this field when querying the bank_admin endpoint. A new Log record
is created whenever a transaction is reconciled, which is done
by a patch request to the bank_admin transaction endpoint.


ibrechin:
@ushkarev ready for review

PR #48, opened 2015-10-19
Update permissions for bank admin updating transactions

ibrechin:
This change allows all bank admins to make updates to transactions,
to allow for the need for non-refund bank admins to update transactions
that have been reconciled.


ibrechin:
@ushkarev please review

PR #49, opened 2015-10-21
Transaction history

ushkarev:
adds additional filters to transactions/payments API and allows it to respond with more details on the transactions

[Partial #104090936]


ushkarev:
@ibrechin could you pls review?

PR #51, opened 2015-10-30
File download tracking

ibrechin:
This is to:

a) enable someone to re-download a copy of the same file (or rather a file including the same set of transactions)

b) track which transactions have already been downloaded for the purpose of creating files from 'all matching transactions which have not yet been included in a file of this type'

The latter will acquire additional work in the API to implement but I'm trying to do smaller PRs.

I might also retire the 'processed' flag as a result of this change, as its sole functionality would be accomplished by b).

### DIFF
--- a/mtp_api/apps/account/models.py
+++ b/mtp_api/apps/account/models.py
@@ -1,0 +1,25 @@
+from django.db import models
+from model_utils.models import TimeStampedModel
+
+from transaction.models import Transaction
+
+
+class FileType(models.Model):
+    name = models.CharField(max_length=100)
+
+    def __str__(self):
+        return self.name
+
+
+class File(TimeStampedModel):
+    file_type = models.ForeignKey(FileType)
+    transactions = models.ManyToManyField(Transaction)
+
+    def __str__(self):
+        return '%s %s' % (self.file_type, self.created)
+
+
+class Balance(TimeStampedModel):
+    opening_balance = models.IntegerField()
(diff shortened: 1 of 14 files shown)


mtp_api/apps/account/models.py
```diff
@@ -0,0 +1,25 @@
+from django.db import models
+from model_utils.models import TimeStampedModel
+
+from transaction.models import Transaction
+
+
+class FileType(models.Model):
```
ushkarev: regarding file types: do you think it's resilient enough to assume that FileType pks are invariant because they're loaded in a fixture? they're used as the identifier in API requests, but do you think that lookups from client apps would need to be done each time a File object needs to be saved?

if the file type set is very rarely going to change, what're your thoughts on having `File.file_type` being a CharField with pre-defined choices?

ibrechin: I think I would prefer to have it defined in a DB table, just as a kind of niceness thing, but I did consider the possibility about accepting the string name in the API and doing a custom lookup based on that. Thoughts?

ushkarev: i suppose `FileType` could have a text code as the primary key so that it's not auto-increment dependent? i'm just not totally sure it warrants a whole model/serialiser/view/url collection…

ibrechin:
I changed to a string PK for file type. I am still in favour of having the types listable via the API, however. This is just on general principals. If a request format features a choice field, the list of possible choices should be available via the API.

PR #53, opened 2015-11-02
addresses Django 1.8+ deprecations

ushkarev:


ushkarev:
@ibrechin, you spotted a deprecation of mine, which made me think of these… would you pls review?

PR #56, opened 2015-11-02
makes the Docker build collect static files

ushkarev:
needed to run Django admin properly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,5 +27,3 @@ venv/
 
 # Node Stuff
 node_modules
-
-assets-src/bower
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 *.pyc
 include/
 lib/
-staticfiles/
+static/
 media/
 .build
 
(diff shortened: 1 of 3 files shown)


ushkarev:
@SteveMarshall could you pls take a look?

SteveMarshall:
Do you mean `mtp_api/assets/.gitignore`, rather than `.gitkeep`? Using `.gitignore` feels like it obscures your intent, and could look accidental… oh, or are you using it to say “keep this directory, but ignore it”?

I might argue that we should add `mtp_api/assets` to the global `.gitignore` and make sure commands don't fail if the directory doesn't  exist: that seems somewhat more intentional and traceable to me. I'm also not sure how many people would think to look anywhere but the root `.gitignore` if a directory seems not to be in git?

ushkarev:
well .gitkeep isn't a real git feature so i've always preferred to use .gitignore as that can then have sensible stuff put into it if needed. i can swap it if you want?

PR #57, opened 2015-11-03
Ensure users' names generated for random test data are more consistent

ushkarev:
to allow for integration tests to expect more specific details (currently login cucumber test in cashbook fails because the name varies)


SteveMarshall:
Great PR! :bowtie:

PR #59, opened 2015-11-04
Download new transactions

ibrechin:
Replaces 'reconciled' field functionality with that provided by file records.

Related to https://github.com/ministryofjustice/money-to-prisoners-bank-admin/pull/28


SteveMarshall:
As we discussed, this probably needs some sort of audit trail for creation of batches and so on, but let's add that as a separate piece of work

PR #64, opened 2015-11-10
Prevent brute force login attacks by locking user accounts

ushkarev:
for half an hour if there are 5 failed attempts ever

### DIFF
--- a/mtp_api/apps/mtp_auth/models.py
+++ b/mtp_api/apps/mtp_auth/models.py
@@ -1,5 +1,8 @@
-from django.db import models
+import datetime
 
+from django.conf import settings
+from django.db import models
+from django.utils.timezone import now
 from model_utils.models import TimeStampedModel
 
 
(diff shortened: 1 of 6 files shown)


mtp_api/apps/mtp_auth/models.py
```diff
@@ -19,3 +22,43 @@ class ApplicationUserMapping(TimeStampedModel):
...
+            user=user,
+            application=client,
+        ).delete()
+
+    @classmethod
+    def add_failed_attempt(cls, user, client):
+        FailedLoginAttempt.objects.create(
+            user=user,
+            application=client,
+        )
```
ibrechin: I think it would be better to have these as methods on the Manager object rather than as class methods on the Model. Just kind of tidier.

mtp_api/apps/mtp_auth/validators.py
```diff
@@ -1,16 +1,31 @@
...
+                raise LockedOut
+
+            valid = super().validate_user(
+                username, password, client, request, *args, **kwargs
+            )
+            if valid and ApplicationUserMapping.objects.filter(
+                    user=request.user, application=client).exists():
+                FailedLoginAttempt.delete_failed_attempts(user, client)
+                return True
+            elif user:
```
ibrechin: What are your thoughts on locking by IP address as well as by username? The typical concern is that someone might choose X common passwords and then just bruteforce on the username instead of the password if one is just trying to get a valid log in. We probably won't have enough users for this to be a serious issue, though.

However, in the current implementation, as it only blocks on valid users, an attacker could brute force valid usernames by checking to see if they get locked out after 5 attempts.

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -146,3 +149,82 @@ def test_bank_admin_cannot_login_to_cashbook(self):
...
+
+        self.assertFalse(FailedLoginAttempt.is_locked_out(prison_clerk, cashbook_client))
+        self.pass_login(prison_clerk, cashbook_client)
+        self.assertFalse(FailedLoginAttempt.is_locked_out(prison_clerk, cashbook_client))
+        self.assertEqual(FailedLoginAttempt.objects.filter(
+            user=prison_clerk,
+            application=cashbook_client,
+        ).count(), 0)
+
+    def test_per_application_account_lockout(self):
```
SteveMarshall: This might be clearer about the intent of the test as `test_lockout_only_applies_to_current_application` or something along those lines?

SteveMarshall: Also, how does the application behave if I get locked out on one application then log in successfully on another? That feels like something else that should be tested (as another test)?

mtp_api/apps/mtp_auth/models.py
ushkarev: yeah i thought that but wasn't sure a whole new class was really necessary

ibrechin: Well, a class definition + setting the objects attribute is only 2 lines and you save 3 by removing the classmethod decorators so I'd say you come out 1 ahead :)

mtp_api/apps/mtp_auth/validators.py
ushkarev: unfortunately, i don't think the user's IP can be passed on to the api app from the client app as part the oauth2 token request, unless we do some hacking and thereby potentially depart from the oauth2 standard…

similarly, without departing from the oauth2 standard, we cannot actually report to the end user that they are locked out directly as part of the token request. it'd still be possible to have a separate api method to determine whether a lockout exists, but then we'd likely be revealing whether a username exists or not…

PR #67, opened 2015-11-12
Update Django and normalise settings across apps

ushkarev:


SteveMarshall:
Retest this please

mtp_api/settings/base.py
```diff
@@ -73,23 +65,29 @@
...
+        'HOST': os.environ.get('DB_HOST', ''),  # Empty for localhost through domain sockets or '127.0.0.1' for localhost through TCP.
+        'PORT': os.environ.get('DB_PORT', ''),  # Set to empty string for default.
+    }
+}
 
-TIME_ZONE = 'UTC'
 
+# Internationalization
+LANGUAGE_CODE = 'en-gb'
+TIME_ZONE = 'UTC'
```
SteveMarshall: Should we be using `Europe/London`, so we don't have to worry about BST changeovers making the times of actions confusing for our users?

mtp_api/settings/base.py
```diff
@@ -1,32 +1,26 @@
...
-https://docs.djangoproject.com/en/1.7/ref/settings/
+https://docs.djangoproject.com/en/1.8/ref/settings/
 """
 import os
 import sys
 
-# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(__file__))
-
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
SteveMarshall: In the other apps, we're doing `from os.path import abspath, dirname, join` and then `dirname(dirname())`. Worth switching this for consistency? (I'm not fussed either way…)

PR #68, opened 2015-11-17
Run clean-up tasks periodically using uWSGI cron facility

ushkarev:
At 3am daily, clears out expired Django sessions and OAuth2 access/refresh tokens


mtp_api/settings/base.py
```diff
@@ -169,6 +169,11 @@
 MTP_AUTH_LOCKOUT_COUNT = 5  # 5 times
 MTP_AUTH_LOCKOUT_LOCKOUT_PERIOD = 30 * 60  # 30 minutes
 
+
+RUN_CLEANUP_TASKS = os.environ.get('RUN_CLEANUP_TASKS') == 'True' or \
+    ENVIRONMENT == 'local'
```
ibrechin: I don't think this should always be true on environment == 'local'. I think that if you want to set something locally, it should be in your settings/local.py.

PR #74, opened 2015-11-26
Add an additional cashbook view for locked credits

ushkarev:


mtp_api/apps/transaction/api/cashbook/serializers.py
```diff
@@ -44,3 +44,14 @@ class Meta:
...
+
+class LockedTransactionSerializer(TransactionSerializer):
+    locked = serializers.BooleanField(read_only=True)
+    locked_at = serializers.DateTimeField(read_only=True)
+
+    class Meta(TransactionSerializer.Meta):
+        fields = TransactionSerializer.Meta.fields + (
+            'locked',
+            'locked_at',
+        )
```
SteveMarshall: Is there a particular reason to only showing these fields for the locked view? Might be that the overall list could show that status too…

mtp_api/apps/transaction/api/cashbook/urls.py
```diff
@@ -5,6 +5,7 @@
 
 urlpatterns = [
     url(r'^transactions/$', csrf_exempt(views.TransactionList.as_view()), name='transaction-list'),
+    url(r'^transactions/locked/$', views.LockedTransactionList.as_view(), name='transaction-locked'),
```
SteveMarshall: Might've been nice to see this as `transactions/?locked=true`, and just another filter on the overall transactions list.

SteveMarshall:
@ushkarev and I discussed my comments this morning. We're going to leave the endpoints as they are because of this PR, but I'd still like to see this functionality migrate into being a filter on the base `transactions` endpoint (maybe ultimately supported by `?group-by` query params and so on later. (Though I know @ushkarev disagrees with me because it requires an extra join for every date-of-action field we add.)

PR #78, opened 2015-12-01
Handle debits

ibrechin:
Handle the upload of debit transactions, which do not need to be processed
by prison staff but which are returned in the bank statement.

Also deal with generating a daily unique ref code for credits.

### DIFF
--- a/mtp_api/apps/transaction/models.py
+++ b/mtp_api/apps/transaction/models.py
@@ -1,13 +1,12 @@
+from datetime import timedelta
 import warnings
 
 from django.db import models
 from django.conf import settings
 from django.dispatch import receiver
-
 from model_utils.models import TimeStampedModel
 
 from prison.models import Prison
-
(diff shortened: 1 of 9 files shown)


mtp_api/apps/transaction/models.py
```diff
@@ -34,30 +34,48 @@ class Transaction(TimeStampedModel):
...
     # set when a transaction is locked and unset if it gets unlocked.
     owner = models.ForeignKey(settings.AUTH_USER_MODEL, null=True, blank=True)
 
     credited = models.BooleanField(default=False)
-
     refunded = models.BooleanField(default=False)
-
     reconciled = models.BooleanField(default=False)
 
     # NB: there are matching boolean fields or properties on the model instance for each
```
ushkarev: should we also update the matching model properties for `locked`, `available` and `refund_pending`?

PR #85, opened 2015-12-15
Include 'debit' transactions in random test data

ibrechin:


ibrechin:
retest this please

mtp_api/apps/transaction/tests/utils.py
```diff
@@ -48,6 +48,7 @@ def generate_initial_transactions_data(tot=50, prisoner_location_generator=None)
         # which is again arbitrary.
         include_prisoner_info = transaction_counter % 5 != 0
         include_sender_roll_number = transaction_counter % 10 == 0
+        make_debit_transaction = (transaction_counter + 1) % 5 == 0
```
ushkarev: is it as many as 1 in 5 because that's how many don't get prisoner info and should get refunded?

ibrechin: yes

PR #88, opened 2015-12-17
Remove 'reset_db' docker task

ibrechin:
This functionality has been replaced by admin management
commands.


ushkarev:
in that case, we should also remove the 'reset_db' option from the `task` choices parameter in a whole bunch of jenkins jobs. not that it's presence would break anything, it'd just be weird to have an option that does nothing

PR #89, opened 2015-12-21
Force Jenkins to run tests [WIP]

SteveMarshall:
This commit is simply to see if Jenkins breaks GitHub builds on test
failures or not, and should never end up on the master branch.


SteveMarshall:
This failed, which means the test passed successfully :boom:

PR #92, opened 2016-01-06
Track invalid references

ibrechin:


mtp_api/apps/transaction/admin.py
```diff
@@ -36,5 +37,22 @@ def display_total_amount(self, request, queryset):
         total = sum(map(lambda t: t.amount, queryset))
         self.message_user(request, 'Total: £%0.2f' % (total / 100))
 
+    def display_reference_validity(self, request, queryset):
+        invalid_ref_count = len(list(filter(lambda t: t.prison is None, queryset)))
+        invalid_percent = (invalid_ref_count / len(queryset)) * 100
```
ushkarev: i'd be careful of `len(queryset)` being 0, you'd get a divide by zero error

mtp_api/apps/transaction/admin.py
```diff
@@ -36,5 +37,22 @@ def display_total_amount(self, request, queryset):
         total = sum(map(lambda t: t.amount, queryset))
         self.message_user(request, 'Total: £%0.2f' % (total / 100))
 
+    def display_reference_validity(self, request, queryset):
+        invalid_ref_count = len(list(filter(lambda t: t.prison is None, queryset)))
```
ushkarev: why not something like `queryset.filter(prison__isnull=True).count()`? that way it doesn't need to load every transaction model and prison related model.

mtp_api/apps/transaction/admin.py
ibrechin: I did miss that, but I think it's ok, as Django prevents you from performing actions if there are no items selected.

PR #97, opened 2016-01-08
Add logging at key points in api transactions

ushkarev:
… which are not already covered by client applications

### DIFF
--- a/mtp_api/apps/transaction/api/bank_admin/views.py
+++ b/mtp_api/apps/transaction/api/bank_admin/views.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+import logging
 
 from django.db.transaction import atomic
 from django.utils import timezone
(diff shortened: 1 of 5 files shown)


mtp_api/apps/transaction/api/bank_admin/views.py
```diff
@@ -66,12 +69,14 @@ def patch_processed(self, request, *args, **kwargs):
         try:
             return self.partial_update(request, *args, **kwargs)
         except Transaction.DoesNotExist as e:
+            transaction_ids = sorted(e.args[0])
+            logger.warn('Some transactions failed to update: [%s]' % ', '.join(map(str, transaction_ids)))
```
ibrechin: There are similar scenarios in `CreditTransactions` and `UnlockTransactions` in api/cashbook/views.py which it might also be worth logging, if you're logging it here.

PR #101, opened 2016-01-14
Integrate new "send money" app

ushkarev:


mtp_api/apps/prison/views.py
```diff
@@ -32,3 +37,44 @@ def create(self, request, *args, **kwargs):
...
+        ActionsBasedPermissions,
+    )
+    serializer_class = PrisonerValiditySerializer
+    lookup_url_kwarg = 'prisoner_reference'
+
+    re_reference = re.compile(r'^(?P<prisoner_number>[a-z]\d{4}[a-z]{2})-'
+                              r'(?P<prisoner_dob>\d{4}-\d{2}-\d{2})$', re.IGNORECASE)
+
+    @classmethod
+    def parse_reference(cls, reference):
```
ibrechin: I don't feel like we need to or should be parsing a reference on this side. The send-money application should have a prisoner number and dob which it then makes a GET request with and sees if it gets something back.

ushkarev: this was supposed to be a rough and ready first attempt. i was going to update this to a POST with two parameters in the future… but i suppose i could do it now, just for you

mtp_api/apps/prison/models.py
```diff
@@ -23,9 +23,12 @@ class PrisonerLocation(TimeStampedModel):
...
 
     class Meta:
-        index_together = [
-            ["prisoner_number", "prisoner_dob"],
-        ]
+        index_together = (
+            ('prisoner_number', 'prisoner_dob'),
+        )
+        permissions = (
+            ('identify_prisoner', 'Can identify if a prisoner exists'),
```
ibrechin: I can't see that you define what this permission means anywhere. In other places this is done by creating an ActionsBasedPermissions subclass which links the permission to a view method.

ibrechin:
No tests?

mtp_api/apps/prison/models.py
ushkarev: actually, action-based permissions aren't needed as there's only one method and the api user is shared for all calls from the send-money app

mtp_api/apps/prison/views.py
```diff
@@ -32,3 +34,29 @@ def create(self, request, *args, **kwargs):
...
+
+    def list(self, request, *args, **kwargs):
+        prisoner_number = self.request.GET.get('prisoner_number', '')
+        prisoner_dob = self.request.GET.get('prisoner_dob', '')
+        try:
+            prisoner_dob = parse_date(prisoner_dob)
+        except ValueError:
+            prisoner_dob = None
+        if not prisoner_number or not prisoner_dob:
+            return Response(None, status=status.HTTP_400_BAD_REQUEST)
```
ibrechin: I think we should return an error message here. Just something saying 'prisoner_dob is required' and 'prisoner_number is required', in the same format as that used for the conflict errors in the cashbook/bank-admin APIs.

PR #103, opened 2016-01-19
Allow for the creation of GOV.UK Pay transactions

ibrechin:
This allows for the creation and processing of transactions
appropriate for integrating the GOV.UK platform. There is a
new category for such payments, and it is possible to mark the
current status of the payment, such that it will not be processed
until the payment has actually been taken.


ibrechin:
So in the end I didn't do any of the discussed refactoring. My reasoning for this is partially that as we don't yet fully understand the GOV.UK payment process, we should wait until we know exactly what needs to be supported before rebuilding everything, and partially that I think it is maybe better not to refactor at the same time as adding new functionality.

PR #110, opened 2016-01-28
Refactor transaction categories

ibrechin:
This returns category to being a binary choice (credit/debit) and
instead stores information about where the transaction came from in
the source field.


mtp_api/apps/transaction/models.py
```diff
@@ -27,6 +27,7 @@ class Transaction(TimeStampedModel):
 
     amount = models.PositiveIntegerField()
     category = models.CharField(max_length=50, choices=TRANSACTION_CATEGORY)
+    source = models.CharField(max_length=50, choices=TRANSACTION_SOURCE)
```
ushkarev: if there's no default, how do existing databases migrate?

ushkarev: oh you specified it directly in the migration

PR #117, opened 2016-02-18
Change password

ibrechin:


mtp_api/apps/mtp_auth/views.py
```diff
@@ -25,3 +28,36 @@ def get_object(self, *args, **kwargs):
...
+class ChangePasswordView(generics.GenericAPIView):
+    permission_classes = (IsAuthenticated,)
+    serializer_class = ChangePasswordSerializer
+
+    @atomic
+    def post(self, request):
+        serializer = self.get_serializer(data=request.data)
+        if serializer.is_valid():
+            old_password = serializer.validated_data['old_password']
+            new_password = serializer.validated_data['new_password']
```
ushkarev: i know you use `password_validation.validate_password` in the client app form, but i think most password validators won't run at the api level on save (which only calls `password_changed` method of validators).

PR #119, opened 2016-02-22
Add field to transaction model to mark lack of sender information

ibrechin:
This is to mark those transactions which, if they cannot be paid to
a prisoner, can also not be easily refunded and may need to be
manually reconciled.


ushkarev:
could this do with a simple test? maybe to make sure the serialisers work as expected?

PR #125, opened 2016-03-08
Prevent reconciliation events being logged multiple times

ibrechin:
Previously, each time a reconciliation file was downloaded,
another reconciliation log record was created for each transaction
in the file.


ibrechin:
retest this please

ushkarev:
good idea! the logs were quite noisy this morning

PR #132, opened 2016-03-23
User admin

ibrechin:
API support for user account management.

Supports:

- CRUD operations on users of the same application and, for cashbook, overlapping prisons
- A user admin can make other user admins of the same application (and prisons)


mtp_api/apps/mtp_auth/models.py
```diff
@@ -21,7 +32,16 @@ class ApplicationUserMapping(TimeStampedModel):
     application = models.ForeignKey('oauth2_provider.Application')
 
     def __str__(self):
-        return self.user.username
+        return '%s -> %s' % (self.user.username, self.application.client_id)
+
+
+class ApplicationGroupMapping(TimeStampedModel):
```
ushkarev: do you think we ought to make a fixture for the default set? or a data migration (though that'd interfere with your tests)

ibrechin: i didn't on the basis that there is no fixture for the default set of applications

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -74,30 +79,332 @@ def test_can_access_my_data_including_managing_prisons(self):
...
+        response = self.client.get(
+            self.get_url(),
+            format='json',
+            HTTP_AUTHORIZATION=self.get_http_authorization_for_user(requester)
+        )
+
+        for user_item in response.data['results']:
+            user = User.objects.get(username=user_item['username'])
+            self.assertEqual(
+                user.applicationusermapping_set.all()[0].application.client_id,
```
ushkarev: not really an issue in this test, but it's possible that some users will be accessing multiple applications (e.g. sharon going into bank and prisoner admins) so it doesn't have to be the first that matches.

ps: you can do `.first()` instead of `.all()[0]`

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -74,30 +79,412 @@ def test_can_access_my_data_including_managing_prisons(self):
...
```
ushkarev: not really an issue in this test, but it's possible that some users will be accessing multiple applications (e.g. sharon going into bank and prisoner admins) so it doesn't have to be the first that matches.

ps: you can do `.first()` instead of `.all()[0]`

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -74,30 +79,412 @@ def test_can_access_my_data_including_managing_prisons(self):
...
+            self.get_url(),
+            format='json',
+            data=user_data,
+            HTTP_AUTHORIZATION=self.get_http_authorization_for_user(requester)
+        )
+
+        make_user_admin = user_data.pop('user_admin', False)
+        new_user = User.objects.get(**user_data)
+        self.assertEqual(
+            new_user.applicationusermapping_set.all()[0].application.client_id,
```
ushkarev: see above

mtp_api/apps/mtp_auth/models.py
ushkarev: i wish we did, even if they don't load automatically it's useful as documentation cos all the pieces are hard to find/remember

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -74,30 +79,332 @@ def test_can_access_my_data_including_managing_prisons(self):
...
+        )
+        self.assertEqual(
+            list(new_user.groups.all()),
+            groups
+        )
+        if hasattr(requester, 'prisonusermapping'):
+            self.assertEqual(
+                list(new_user.prisonusermapping.prisons.all()),
+                list(requester.prisonusermapping.prisons.all())
+            )
```
ushkarev: again, not so much of an issue for tests yet, but the mappings have no specific order so comparing sets (of prison-user pairs) would work better than lists (of models)

mtp_api/apps/mtp_auth/models.py
ibrechin: indeed, don't think could auto load applications as they aren't entirely repeatable (user/secret)

i see your point in that it might be nice if at least for the tests they were loaded from a fixture rather than generated by code...although again there's the user link requirement...i guess we'd need a basic user fixture as well

certainly i'm not opposed to the idea but i'd prefer to leave it out of this PR

PR #136, opened 2016-03-30
API upgrades

ushkarev:


ushkarev:
@ibrechin i was just thinking it's good to periodically to update dependencies… you cool with the idea in principle?

however there's an annoying issue – django-oauth-toolkit decided to split a migration in two in a non-backwards compatible way. so we would need to manually add this to all databases (including our local versions):

    INSERT INTO "django_migrations"("app","name","applied")
    VALUES
    ('oauth2_provider', '0002_08_updates', NOW());

PR #146, opened 2016-04-13
Admin dashboard for basic stats overiew

ushkarev:


mtp_api/apps/core/views.py
```diff
@@ -6,39 +6,86 @@
...
+class AdminViewMixin:
     """
-    Django admin view which calls load_test_data management command
+    Mixin for custom MTP django admin views
     """
-    form_class = RecreateTestDataForm
-    template_name = 'core/recreate_test_data.html'
-    success_url = reverse_lazy('mtp-admin:recreate_test_data')
+    disable_in_production = False
+    superuser_required = False
```
ibrechin: Should these be overridden in ```RecreateTestDataView```?

mtp_api/apps/core/dashboards.py
```diff
@@ -0,0 +1,224 @@
...
+
+    class Media:
+        css = {
+            'all': ('core/css/transaction-report.css',)
+        }
+
+    def __init__(self, received_at=None):
+        if not received_at:
+            self.title = _('Today’s transaction report')
+            received_at = now().date()
```
ibrechin: Should the default rather be yesterday, as this is when the most recent transactions will be dated?

mtp_api/apps/mtp_auth/views.py
```diff
@@ -6,7 +6,7 @@
 from django.db.transaction import atomic
 from django.forms import ValidationError
 from django.http import Http404
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import gettext_lazy as _
```
ibrechin: Por que?

mtp_api/apps/core/views.py
ushkarev: quite right

mtp_api/apps/core/dashboards.py
ushkarev: quite right, i was being stupid

mtp_api/apps/mtp_auth/views.py
ushkarev: parce que le python 3

ibrechin: bueno

PR #159, opened 2016-04-27
Fix JS bug for setting cookies for a dashboard "stand out" module

ushkarev:


mtp_api/assets/core/js/dashboard.js
```diff
@@ -8,6 +8,7 @@ django.jQuery(function($) {
     cookieName = $dashboardWrapper.data('cookie-name'),
     standoutCookieName = $dashboardWrapper.data('standout-cookie-name'),
     $moduleForms = $('.mtp-dashboard-change form'),
+    $standoutModule = $('#' + Cookies.get(standoutCookieName)),
     autoreloadInterval;
```
maxf: Optionally you could use the $.cookie plugin to remain in the jquery namespace

maxf:
I still think this mode should be available through a URL param rather than a cookie

PR #169, opened 2016-05-06
Refactor credits

ibrechin:
Create a 'Credit' model which will be used to represent a payment which due a
prisoner. Credits can be attached to either Transactions (in the case of bank 
transfers) or Payments (in the case of debit card payments).

The client APIs are largely unchanged, but for the endpoints. Cashbook now uses
'/credits' instead of '/cashbook/transactions' and Bank Admin and Transaction
Uploader now use '/transactions' instead of '/bank_admin/transactions'.


mtp_api/apps/mtp_auth/fixtures/initial_groups.json
```diff
@@ -29,10 +29,10 @@
...
     "name": "PrisonClerk",
     "permissions": [
-      ["view_transaction", "transaction", "transaction"],
-      ["lock_transaction", "transaction", "transaction"],
-      ["unlock_transaction", "transaction", "transaction"],
-      ["patch_credited_transaction", "transaction", "transaction"]
+      ["view_credit", "credit", "credit"],
+      ["lock_credit", "credit", "credit"],
+      ["unlock_credit", "credit", "credit"],
+      ["patch_credited_credit", "credit", "credit"]
```
ushkarev: would we manually get these onto prod?

mtp_api/apps/credit/managers.py
```diff
@@ -0,0 +1,82 @@
...
+    def refunded(self):
+        return self.filter(**self.model.STATUS_LOOKUP[CREDIT_STATUS.REFUNDED])
+
+    def refund_pending(self):
+        return self.filter(**self.model.STATUS_LOOKUP[CREDIT_STATUS.REFUND_PENDING])
+
+    def locked_by(self, user):
+        return self.filter(owner=user)
+
+    def update_prisons(self):
```
ushkarev: this doesn't really feel like a query set method because it isn't compose-able with others, perhaps makes sense as a model manager method or even model class method? not too bothered tho

PR #179, opened 2016-05-16
Prevent charts overlaying each other when a module is "standing out" in the admin dashboard

ushkarev:


maxf:
Just a general point: the admin & dashboard code aren't API code per se, and should perhaps be removed from the API to ensure better separation of concerns, to keep the API code being just about the API and the rest of the code as API clients. Also it would prevent the API itself exposing a web interface, ie an attack vector. This is obviously made more difficult by django currently making it effortless to have an admin front-end, but not the best design.

PR #190, opened 2016-05-19
Include debit card payments in generated test data

ibrechin:
This is to ensure that our applications continue to work correctly
when there are debit card payments being made.


ibrechin:
retest this please

PR #191, opened 2016-05-19
Add sorting and more filtering to security APIs

ushkarev:


mtp_api/apps/credit/views.py
```diff
@@ -112,10 +113,14 @@ class CreditListFilter(django_filters.FilterSet):
...
-    sender_account_number = django_filters.CharFilter(name='transaction__sender_account_number')
-    sender_roll_number = django_filters.CharFilter(name='transaction__sender_roll_number')
+    sender_name = django_filters.CharFilter(name='transaction__sender_name',
+                                            lookup_expr='icontains')
+    sender_sort_code = django_filters.CharFilter(name='transaction__sender_sort_code',
+                                                 lookup_expr='icontains')
+    sender_account_number = django_filters.CharFilter(name='transaction__sender_account_number',
+                                                      lookup_expr='icontains')
+    sender_roll_number = django_filters.CharFilter(name='transaction__sender_roll_number',
+                                                   lookup_expr='icontains')
```
ibrechin: I don't think it's worth doing an ```icontains``` on the account details, and I think it will slow down DB queries if we index on these fields (which we may well).

mtp_api/apps/credit/urls.py
```diff
@@ -9,5 +9,5 @@
     url(r'^credits/actions/lock/$', views.LockCredits.as_view(), name='credit-lock'),
     url(r'^credits/actions/unlock/$', views.UnlockCredits.as_view(), name='credit-unlock'),
     url(r'^credits/senders/$', csrf_exempt(views.SenderList.as_view()), name='sender-list'),
-    url(r'^credits/recipients/$', csrf_exempt(views.RecipientList.as_view()), name='recipient-list'),
+    url(r'^credits/prisoners/$', csrf_exempt(views.PrisonerList.as_view()), name='prisoner-list'),
```
ibrechin: There was a reason for calling it 'recipients' rather than 'prisoners' which is that they are not equal sets - although all recipients are prisoners, not all prisoners are recipients. Doesn't really matter, though.

PR #193, opened 2016-05-20
Add filtering on the amount of a credit for intelligence analysis

ibrechin:
This filtering can be on magnitude or on certain patterns such
as the least significant digits of a regex. These two pattern
types can also be excluded from the result set.


ibrechin:
retest this please

PR #194, opened 2016-05-23
Refactor raw SQL generation to avoid direct param insertion

ibrechin:
Return a tuple of query string and param list from Form SQL filter
constructor rather than returning a query string with the params
inserted. Also return a tuple from View SQL filter constructor
to avoid mutation of param list variable.

### DIFF
--- a/mtp_api/apps/credit/forms.py
+++ b/mtp_api/apps/credit/forms.py
@@ -1,5 +1,9 @@
+from collections import namedtuple
+
 from django import forms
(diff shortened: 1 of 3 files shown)


ibrechin:
retest this please

ibrechin:
retest this please

ibrechin:
Refactoring is up for discussion. If you don't like it, can make a separate PR for the bug fixes.

PR #204, opened 2016-05-31
Reject anonymous payments

ibrechin:
Any payment which has incomplete sender information will no longer
be credited to a prisoner, even if it is possible to identify the
prisoner for whom it is intended. Instead it will be classified as
an 'unidentified' payment and no automatic action will be taken.

Payments lacking sender information which have been credited in the
past will still show up as successful payments.


ibrechin:
This should not be merged until process and messaging have been updated.

PR #210, opened 2016-06-03
Display a one week chart on the 'transactions today' dashboard

ibrechin:
Slightly more interesting than displaying nothing.

In order to accomplish this the TransactionReportChart class becomes
responsible for generating its own date-ranged queryset.


ibrechin:
retest this please

PR #211, opened 2016-06-09
Improve satisfaction chart clarity

ushkarev:


mtp_api/assets/core/js/satisfaction-results.js
```diff
@@ -48,27 +53,112 @@ django.jQuery(function($) {
...
         y: legendBounds.top - 6,
         fill: '#000',
-        'font-family': 'Roboto, "Lucida Grande", Verdana, Arial, sans-serif',
+        'font-family': font,
         'font-weight': 'bold',
         'font-size': 13
       });
       $svg.append($title);
+
+      $svg.on('mouseover', function() {
```
maxf: tablets? We don't have proper browser requirements for the admin features, but we should perhaps consider that some day a user will be looking at those numbers from their mobile device.

mtp_api/assets/core/js/satisfaction-results.js
```diff
@@ -48,27 +53,112 @@ django.jQuery(function($) {
...
     });
 
     google.visualization.events.addListener(chart, 'ready', function() {
       var $svg = $chart.find('svg'),
+        svgNamespace = $svg[0].namespaceURI,
         cli = chart.getChartLayoutInterface(),
         legendBounds = cli.getBoundingBox('legend'),
-        $title;
+        chartBounds = cli.getChartAreaBoundingBox(),
+        significantWidth = chartBounds.width * 5 / 6, // last option is disregarded
```
maxf: What does this comment mean? Is it a bug?

mtp_api/assets/core/js/satisfaction-results.js
ushkarev: yeah i know… at a glance, with this PR you'll only see the average response until you mouse-over (then you see the actual responses). if that's sufficient for mobile devices i guess it's ok.

mtp_api/assets/core/js/satisfaction-results.js
ushkarev: it's deliberate, the average response is somewhere between "strongly disagree" and "strongly agree", i.e. "not applicable" (the 6th response) is disregarded for the width calculation.

PR #213, opened 2016-06-16
Admin improvements

ushkarev:


mtp_api/apps/account/admin.py
```diff
@@ -1,17 +1,48 @@
...
+            if transaction.category == TRANSACTION_CATEGORY.CREDIT:
+                params['status'] = 'credit (%s)' % transaction.status
+            else:
+                params['status'] = 'debit'
+            return (
+                reverse('admin:transaction_transaction_change', args=(transaction.id,)),
+                '%(amount)s %(status)s, %(date)s' % params,
+            )
+
+        return format_html_join(mark_safe('<br>\n'), '<a href="{}">{}</a>', (
```
SteveMarshall: `<br>`? 😞 What's the context of this markup? Can we wrap it in something a little more semantic (like a `ul` or paragraphs or something)?

ushkarev: it's the MTP-team-only admin tool, there's no need for semantic markup that would need special styling
screen shot 2016-06-16 at 13 46 14

SteveMarshall: If @maxf is happy with it, then I am too. (And with the rest of this PR, so he can merge!)

maxf: It's part of the Django admin HTML that's not our markup. There'd be a lot to do to fix it (and if we ever want to have proper admin pages, it should be a separate app anyway).

PR #220, opened 2016-06-27
Remove API-side validation that the user can access the client application

ibrechin:
This is now checked by the client application. This is considered to be
sufficient as it makes sense for the client to handle access to itself.
The only additional attack vector that I can think of is if someone had
e.g. a bank admin's user credentials and the client id/secret for a
different app then they could obtain an oauth token, but they could
do the same simply by logging in to the bank admin app, so nbd.


ibrechin:
retest this please

ibrechin:
retest this please

PR #222, opened 2016-06-30
Replace prison gender with population and category

ibrechin:
Population and category encompass the type of prison.


ibrechin:
retest this please

PR #226, opened 2016-07-04
Apply MIT license

SteveMarshall:
This codebase should have always been MIT licensed.


SteveMarshall:
Apparently @maxf already did this in #225!

PR #227, opened 2016-07-05
Optimisations

ibrechin:
* Allow for a batch to be updated after intial creation
  * This means that not all transactions have to be included in the create request, allowing for batching of large numbers of transactions.
* Optimise the reconciliation of transactions
  * Process credits and transactions separately
  * Bulk update credits to reconciled, and bulk create credit reconciliation
logs
  * Use temp table insert to do bulk update of generated ref codes on
transactions

### DIFF
--- a/mtp_api/apps/transaction/managers.py
+++ b/mtp_api/apps/transaction/managers.py
@@ -11,14 +11,15 @@ class TransactionQuerySet(models.QuerySet):
     def reconcile(self, date, user):
         update_set = self.filter(
             received_at__gte=date,
-            received_at__lt=(date + timedelta(days=1))
+            received_at__lt=(date + timedelta(days=1)),
+            credit__isnull=False,
+            credit__reconciled=False
         ).order_by('id').select_for_update()
 
         ref_code = settings.REF_CODE_BASE
         for transaction in update_set:
-            if transaction.credit and not transaction.credit.reconciled:
-                if transaction.reconcilable:
-                    transaction.ref_code = ref_code
-                    ref_code += 1
-                transaction.credit.reconcile(user)
-                transaction.save()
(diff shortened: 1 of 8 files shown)


mtp_api/apps/transaction/managers.py
```diff
@@ -1,25 +1,39 @@
...
+            with connection.cursor() as c:
+                c.execute('DROP TABLE IF EXISTS refids;')
+                c.execute('CREATE TEMP TABLE refids (id integer, ref_code integer)')
+                insert_query = 'INSERT INTO refids (id, ref_code) VALUES '
+                insert_query += ', '.join(['%s' for _ in ref_codes])
+                insert_query += ';'
+                c.execute(insert_query, ref_codes)
+                c.execute('UPDATE transaction_transaction t SET '
+                          'ref_code=r.ref_code FROM refids r WHERE '
+                          't.id=r.id;')
```
SteveMarshall: All of this raw SQL (and particularly the `join` in the `insert_query`) worries me. Is there anything in the ORM we can use instead of hand-crafting these?

SteveMarshall:
Other than the raw SQL, this looks good to me. I might prefer a slightly more descriptive PR title than “Optimisations”, though 😝

mtp_api/apps/transaction/managers.py
ibrechin: Unfortunately not. This is basically the only way we can do a bulk update with varying values. Re: the join, bear in mind that that doesn't actually insert any dynamic data into the query.

SteveMarshall: Oh, totally, I'm just always wary of any raw SQL in code.

@ohmygauss, do you have any opinions on this?

PR #237, opened 2016-08-18
Improve test data generation

ibrechin:


mtp_api/apps/core/management/commands/load_test_data.py
```diff
@@ -33,7 +33,12 @@ def add_arguments(self, parser):
                             help='Prevents existing credits from being deleted')
         parser.add_argument('--prisons', nargs='*', default=['sample'],
                             choices=['sample', 'nomis'],
-                            help='Create prisons from these sets')
+                            help='Create prisons from these sets'),
```
ushkarev: don't think you want a trailing comma here

PR #252, opened 2016-09-16
Avoid setting prison to null on a debit card payment

ibrechin:
As debit card refunds need to be reconciled manually, we cannot track
such refunds within our system. If an update to prisoner locations will
mean that a previously matched debit card payment will no longer be
creditable, do not update that payment's corresponding credit. Instead it
will remain assigned to the last matched prison.


ibrechin:
retest this please

PR #259, opened 2016-09-29
Add command to delete zendesk tickets older than two weeks

ibrechin:
We clead up older, closed tickets in order to avoid keeping
potentially personal information in zendesk longer than strictly
necessary.

### DIFF
--- a/mtp_api/apps/core/management/commands/delete_zendesk_tickets.py
+++ b/mtp_api/apps/core/management/commands/delete_zendesk_tickets.py
@@ -1,0 +1,40 @@
+from datetime import timedelta
+
+from django.conf import settings
+from django.core.management import BaseCommand
+from django.utils import timezone
+import requests
+
+SEARCH_URL = settings.ZENDESK_BASE_URL + '/api/v2/search.json'
+DELETE_URL = settings.ZENDESK_BASE_URL + '/api/v2/tickets/destroy_many.json'
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **kwargs):
+        auth = (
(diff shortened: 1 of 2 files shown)


mtp_api/apps/core/management/commands/delete_zendesk_tickets.py
```diff
@@ -0,0 +1,42 @@
...
+
+        query = 'type:ticket status:closed group_id:{group_id} updated<{two_weeks}'.format(
+            group_id=settings.ZENDESK_GROUP_ID,
+            two_weeks=(timezone.now() - timedelta(weeks=2)).strftime('%Y-%m-%d')
+        )
+        url = SEARCH_URL + '?page=1&query=%s' % query
+
+        while True:
+            run_again = False
+            response = requests.get(url, auth=auth, timeout=15)
```
ushkarev: why not pass `page` and `query` as a dict, that way escaping : and < etc is done automatically

mtp_api/apps/core/management/commands/delete_zendesk_tickets.py
```diff
@@ -0,0 +1,42 @@
...
+
+            if response.status_code == 200:
+                data = response.json()
+                num_results = len(data['results'])
+                if num_results > 0:
+                    delete_url = DELETE_URL + '?ids={ids}'.format(
+                        ids=','.join(map(str, [r['id'] for r in data['results']]))
+                    )
+
+                    requests.delete(delete_url, auth=auth, timeout=15)
```
ushkarev: same as above for `ids`

PR #266, opened 2016-10-11
Produce reliable set of credits for testing ordering by prisoner name

ushkarev:
…because [postgres collates ignoring spaces](https://wiki.postgresql.org/wiki/FAQ#Why_do_my_strings_sort_incorrectly.3F)


ushkarev:
retest this please

PR #271, opened 2016-10-14
Refactor credit processing operations

ibrechin:
- Moved business logic from the views to the manager class.
- Batched all update/create operations
- Use select_for_update to ensure concurrency safety


ibrechin:
This is a subjective one so just like see what you think I guess.

PR #274, opened 2016-10-25
Fix reconciliation bounds

ibrechin:


ibrechin:
retest this please

mtp_api/apps/transaction/migrations/0039_consistent_received_at_time.py
```diff
@@ -0,0 +1,41 @@
...
+
+        received_at = transaction.received_at
+        if received_at.timetz() in (eleven, midnight):
+            # check if recorded under BST
+            if received_at.timetz() == eleven:
+                received_at += timedelta(days=1)
+
+            received_at = datetime.combine(received_at.date(), noon)
+            transaction.received_at = received_at
+            transaction.save()
```
ushkarev: could it possibly have been done with something like this?

```python
Transaction.objects.filter(received_at__time=time(11, tzinfo=utc)).update(received_at=F(received_at) + timedelta(hours=1))
```

but no big deal as only will run once :)

PR #297, opened 2016-11-24
Add a step to get gov.uk template and assets to Makefile

ushkarev:
(to make 404 and 500 backup pages work)


ibrechin:
How about instead we just change the view handler to return an empty response?

ushkarev:
alternative route [api#298](https://github.com/ministryofjustice/money-to-prisoners-api/pull/298)

PR #306, opened 2016-12-14
Refactor security data

ibrechin:
Not really a refactor, it's just an extra thing and thus backwards compatible.


ushkarev:
mtp_api/apps/security/__init__.py (can't comment on it directly cos there are no lines in it) needs: `default_app_config = 'security.app.AppConfig'`

mtp_api/apps/security/management/commands/update_security_profiles.py
```diff
@@ -0,0 +1,107 @@
...
+    PrisonerProfile, SecurityDataUpdate, CardholderName
+)
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **kwargs):
+        try:
+            last_updated_at = SecurityDataUpdate.objects.latest().updated_at
+            new_credits = Credit.objects.filter(received_at__gt=last_updated_at)
```
ushkarev: might be safer to do `received_at__gte`… in case there are a few at the same moment

mtp_api/apps/security/management/commands/update_security_profiles.py
```diff
@@ -0,0 +1,107 @@
...
+        except SecurityDataUpdate.DoesNotExist:
+            new_credits = Credit.objects.all()
+
+        for credit in new_credits:
+            self.create_or_update_profiles(credit)
+
+        if len(new_credits):
```
ushkarev: since you never access `new_credits` contents directly, seems to me that `new_credits.exists()` is more appropriate than loading all of them

mtp_api/apps/security/models.py
```diff
@@ -0,0 +1,97 @@
...
+
+class BankTransferSenderDetails(TimeStampedModel):
+    sender_name = models.CharField(max_length=250, blank=True)
+    sender_sort_code = models.CharField(max_length=50, blank=True)
+    sender_account_number = models.CharField(max_length=50, blank=True)
+    sender_roll_number = models.CharField(max_length=50, blank=True)
+    sender = models.ForeignKey(
+        SenderProfile, on_delete=models.CASCADE, related_name='bank_transfer_details'
+    )
+
```
ushkarev: what makes a sender unique? i'd probs want to make sort-code, account and roll num unique, but not sure what to do about name. especially since in the debit card version, the name is an entirely separate model

mtp_api/apps/security/models.py
```diff
@@ -0,0 +1,97 @@
...
+            models.Q(prisoner_dob=self.prisoner_dob)
+        )
+
+    class Meta:
+        permissions = (
+            ('view_prisonerprofile', 'Can view prisoner profile'),
+        )
+
+
+class SecurityDataUpdate(models.Model):
```
ushkarev: since sender and prisoner profiles are timestamped, doesn't the `.modified` field on it suffice for this purpose? not especially against this method tho.

mtp_api/apps/security/management/commands/update_security_profiles.py
ibrechin: I already have, like two lines up,

mtp_api/apps/security/management/commands/update_security_profiles.py
ibrechin: Yeah, I thought about that. Although in that case you might then get credits counted multiple times. Might change to id.

mtp_api/apps/security/management/commands/update_security_profiles.py
ushkarev: too far away for me to see

mtp_api/apps/security/management/commands/update_security_profiles.py
ushkarev: aha pk might be the way to go… also, at what point do we run out of pks?

mtp_api/apps/security/models.py
ibrechin: But the updated_at is based on the received_at time of the credit, rather than the actual time that the update was run.

mtp_api/apps/security/models.py
ibrechin: i think it should be, because e.g. then we can count metro bank senders separately because although they all have the same sort code/account number they do have different sender names

ushkarev: cool

mtp_api/apps/security/management/commands/update_security_profiles.py
ibrechin: apparently they are signed 32-bit, which frankly is pretty disappointing

PR #308, opened 2016-12-16
Tweak security profiles in admin

ushkarev:


mtp_api/templates/admin/security/prisonerprofile/change_form.html
```diff
@@ -0,0 +1,17 @@
...
+{% load i18n %}
+
+{% block after_related_objects %}
+  <div class="inline-group">
+    <h2>{% trans 'Sender profiles' %}</h2>
+    <ul>
+      {% for sender in original.senders.all %}
+        <li>
+          <a href="{% url 'admin:security_senderprofile_change' sender.pk %}">
+            {{ sender.get_sender_names|join:', ' }}
```
ibrechin: Sender name can be blank. If you want to use sender names here need some kind of "anonymous" default to provide link text.

ushkarev: i know, but thought this was better than "Sender 19321". i'll add another method :)

PR #326, opened 2017-02-08
Add support for storing saved searches for users

ibrechin:
This allows users of the security tool to save a search, pin a profile,
have default filters etc.

Only the creating user may interact with a given saved search via the API.


ibrechin:
So I've got this thing with the `filters` and the `detail_url`

Basically if you're saving a search then only the `filters` are set, whereas if you're pinning a profile then the `detail_url` is the URL of that profile, and then the `filters` are the filters for the credit list on that profile (so that you can still check the number of new credits).

Does that seem right to you?

PR #337, opened 2017-02-27
Update django version to 1.10 and fix associated issues

ibrechin:


requirements/dev.txt
```diff
@@ -1,7 +1,7 @@
 # Place development dependencies here
 -r base.txt
 
-money-to-prisoners-common[testing]>=5.0.4,<5.1
+money-to-prisoners-common[testing]>=5.5,<5.6
 
 django-extensions==1.6.1
 six==1.10.0
```
ushkarev: how about these two lines :)

mtp_api/apps/core/filters.py
```diff
@@ -39,6 +39,23 @@ def filter(self, qs, value):
...
+                        model, name, self.lookup_expr, self.exclude
+                    ))
+                self._label = ', '.join(fields)
+            return self._label
+
+        def fset(self, value):
+            self._label = value
+
+        return locals()
+
```
ushkarev: i don't django filters do this weird thing with property and locals – is it a backwards compatibility thing for python 2.0? is this not equivalent to

```python
    @property
    def label(self):
        if self._label is None and hasattr(self, 'parent'):
            model = self.parent._meta.model
            fields = []
            for name in self.name:
                fields.append(django_filters.utils.label_for_filter(
                    model, name, self.lookup_expr, self.exclude
                ))
            self._label = ', '.join(fields)
        return self._label

    @label.setter
    def label(self, value):
        self._label = value
```

ibrechin: I was mainly just copying what they had with the vague idea that it would make it easier to keep in line with upstream changes but that's probably not really a thing. Have changed to less gross version.

PR #341, opened 2017-03-07
Link credits to security profiles directly to speed up searching

ushkarev:
and make miscounted totals easier to correct


mtp_api/apps/security/views.py
```diff
@@ -98,11 +98,14 @@ class SenderProfileCreditsView(
...
     def list(self, request, sender_pk=None):
         sender = get_object_or_404(SenderProfile, pk=sender_pk)
-        queryset = self.filter_queryset(self.get_queryset())
-        queryset = queryset.filter(sender.credit_filters)
+        queryset = self.get_queryset().filter(sender_profile=sender)
+        queryset = self.filter_queryset(queryset)
 
         page = self.paginate_queryset(queryset)
         if page is not None:
+            if sender.credit_count != self.paginator.count:
```
ushkarev: i don't really like these bits

PR #349, opened 2017-03-21
Updated user account management

ushkarev:


mtp_api/apps/mtp_auth/serializers.py
```diff
@@ -16,8 +16,8 @@
 
 
 class UserSerializer(serializers.ModelSerializer):
-    applications = serializers.SerializerMethodField()
-    permissions = serializers.SerializerMethodField()
+    applications = serializers.SerializerMethodField(read_only=True)
+    permissions = serializers.SerializerMethodField(read_only=True)
```
ibrechin: `SerializerMethodField` is read-only by nature. Or is something complaining because it is not explicitly set?

ushkarev: yeah they are read-only, but i just wanted to make it a little clearer which fields are in fact used as POST parameters and which aren't. happy to remove if you like

ibrechin: i find it kind of confusing

mtp_api/apps/mtp_auth/serializers.py
```diff
@@ -86,67 +99,103 @@ def get_prisons(self, obj):
...
             pu = PrisonUserMapping.objects.create(user=new_user)
             for prison in prisons:
                 pu.prisons.add(prison)
-            pu.save()
+
+        main_application_name = None
+        for role in roles:
+            if main_application_name is None:
+                main_application_name = role.application.name
+            ApplicationUserMapping.objects.create(user=new_user, application=role.application)
```
ibrechin: Would this create multiple identical mappings where the user has multiple roles with access to the same application?

ushkarev: this is the create method so the user did not exist before now

mtp_api/apps/mtp_auth/serializers.py
```diff
@@ -86,67 +99,103 @@ def get_prisons(self, obj):
...
         updated_user = super().update(user, validated_data)
 
-        if make_user_admin is not None:
-            user_admin_group = Group.objects.get(name='UserAdmin')
+        if roles:
+            updated_user.applicationusermapping_set.all().delete()
+            updated_user.groups.clear()
+
+            if make_user_admin:
+                updated_user.groups.add(user_admin_group)
```
ibrechin: Why not have user admin as a role, independent of any particular application?

ushkarev: the idea is that a user is only ever in one role officially. that way the form choices are simple and must be linked to an application

ushkarev:
retest this please

mtp_api/apps/mtp_auth/views.py
```diff
@@ -35,19 +47,31 @@ class UserViewSet(viewsets.ModelViewSet):
...
+        Set is determined by matching set of prisons if some exist OR by key group as determined by Role models.
+        If a user falls into multiple roles, they can only edit themselves.
+        """
+        user = self.request.user
+        queryset = User.objects.filter(is_superuser=user.is_superuser).order_by('username')
+
+        try:
+            prisons = list(PrisonUserMapping.objects.get(user=user).prisons.values_list('pk', flat=True))
+        except PrisonUserMapping.DoesNotExist:
+            prisons = []
```
ibrechin: Could use `PrisonUserMapping.objects.get_prison_set_for_user` to avoid having to handle `DoesNotExist` here.

mtp_api/apps/mtp_auth/views.py
```diff
@@ -200,7 +224,7 @@ def post(self, request):
 
             send_email(
                 user.email, 'mtp_auth/reset_password.txt',
-                gettext('Your new Money To Prisoners password'),
+                gettext('Your new Prisoner Money password'),
```
ibrechin: Has Florian or someone officially decided that we're changing the user facing instances of the service name or is this you taking the initiative?

ushkarev: this one was me. but it's not like we use "Money To Prisoners" anywhere else ever

mtp_api/apps/mtp_auth/serializers.py
ibrechin: Sure, it's an edge case, but it is technically possible. Could maybe instead get the list of applications via a set comprehension? I dunno, might not be enough of a big deal.

mtp_api/apps/mtp_auth/serializers.py
ibrechin: in that case should it be possible (as it is) to create/update a user such that they have multiple roles?

ushkarev:
retest this please

mtp_api/apps/mtp_auth/serializers.py
ushkarev: the api allows it but the forms don't let you edit those people

PR #354, opened 2017-03-27
Add nomis integration flag to prisoner validity check

ibrechin:
This is to inform send-money about different behaviour that needs
to be used for prisoners who are in prisons where NOMIS integration
has been rolled out.


ibrechin:
This is for deciding what copy to use in emails/confirmation page (i.e. do we tell them that they will receive an email when the prisoner is credited).

PR #392, opened 2017-09-06
Catch errors for scheduled commands

ibrechin:
This will prevent a failing scheduled command from stopping others
to run.


ibrechin:
retest this please

PR #395, opened 2017-10-06
Fix credit date selection for prisoner credit notices generation

ushkarev:


ibrechin:
Would it perhaps be better to explicitly set the timezone to UTC, rather than make_aware?

ushkarev:
not a big deal, imo. nobody will be crediting around midnight. made sense logically to use local timezone for something done by humans

PR #398, opened 2017-11-10
Add API support for disbursements

ibrechin:
Introduces new models for disbursement and recipient, and API
views for creation and updating of these new models.


mtp_api/apps/disbursement/tests/test_views.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class CreateDisbursementTestCase(AuthTestCaseMixin, APITestCase):
+    fixtures = ['initial_types.json', 'test_prisons.json', 'initial_groups.json']
+
+    def setUp(self):
+        super().setUp()
+        self.prison_clerks, _, self.bank_admins, _, _, _ = make_test_users()
+
+    def test_permissions_required(self):
```
ushkarev: i think it'd be nice to have a test for a prison clerk failing to make a disbursement for someone not in their prison… or do you feel that we don't need to force that rule?

mtp_api/apps/disbursement/serializers.py
```diff
@@ -0,0 +1,40 @@
...
+class RecipientSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Recipient
+        fields = '__all__'
+
+
+class DisbursementSerializer(serializers.ModelSerializer):
+    @atomic
+    def create(self, *args, **kwargs):
+        new_disbursement = super().create(*args, **kwargs)
```
ushkarev: how will we try to re-use existing models?

mtp_api/apps/disbursement/serializers.py
```diff
@@ -0,0 +1,40 @@
+from django.db.transaction import atomic
+from rest_framework import serializers
+
+from .models import Disbursement, Recipient
+from .signals import disbursement_created
+
+
+class RecipientSerializer(serializers.ModelSerializer):
```
ushkarev: how will we try to re-use existing models?

ibrechin: User would get list of recipients filtered by prisoner/prison/whatever (not implemented because it hasn't actually been designed in), then supply chosen id when creating a new disbursement.

mtp_api/apps/disbursement/tests/test_views.py
ibrechin: I guess probably should for consistency.

mtp_api/apps/disbursement/serializers.py
```diff
@@ -0,0 +1,80 @@
...
+        prison = data['prison']
+
+        try:
+            PrisonerLocation.objects.get(
+                prisoner_number=prisoner_number,
+                prison=prison
+            )
+        except PrisonerLocation.DoesNotExist:
+            raise serializers.ValidationError(
+                _('Prisoner {prisoner_number} is not in {prison}').format(
```
ushkarev: Transifex doesn't work with new format strings

PR #407, opened 2017-12-04
Create a management command to synchronise user group permissions with canonical fixture

ushkarev:


ibrechin:
Maybe some tests?

PR #411, opened 2017-12-07
Stabilise notification level ordering

ushkarev:
it was previously dependent on volatile dictionary order


ibrechin:
retest this please

PR #419, opened 2018-01-10
Allow disbursement reset

ibrechin:


ibrechin:
i didn't bother fixing the exceptions because it's in your branch already

PR #426, opened 2018-01-30
Change transaction type code for digital take-up report

ushkarev:


ibrechin:
:rotating_light: WEEOOO WEEOOO :rotating_light:

PR #432, opened 2018-02-14
Add invoice number to disbursements

ushkarev:
which gets prefixed with "PMD" starting from id 100 for easier searching in AP


ushkarev:
retest this please

PR #453, opened 2018-05-16
Fix slow queries when getting digital takeup for prison performance

ibrechin:


ibrechin:
Here, I basically copied what you did but with a few subtle differences to try and justify my existence.

ushkarev:
ha well given that i did it wrong, that's all good ;)

PR #454, opened 2018-05-16
Digital take-up and savings report

ushkarev:


mtp_api/apps/performance/models.py
```diff
@@ -32,6 +33,34 @@ def digital_takeup_per_day(self):
...
+                    SUM(credits_by_post) AS reported_credits_by_post,
+                    SUM(credits_by_mtp) AS reported_credits_by_mtp
+                  FROM performance_digitaltakeup
+                  WHERE date >= %(min_date)s
+                  GROUP BY date_trunc('month', date)
+                )
+                SELECT credit_count.date, accurate_credits_by_mtp, reported_credits_by_post, reported_credits_by_mtp
+                FROM credit_count FULL OUTER JOIN average_takeup ON credit_count.date = average_takeup.date
+                ORDER BY credit_count.date
+            ''', params={'min_date': datetime.date(2017, 1, 1)})
```
ibrechin: Is having this as the only minimum sustainable? Shouldn't it also be based on the period that you're looking at?

mtp_api/apps/performance/views.py
```diff
@@ -150,18 +156,18 @@ def get_context_data(self, **kwargs):
...
+            reverse=bool(form.cleaned_data['desc'])
         )
 
+        query_prefix = '&'.join(
+            '%s=%s' % (name, value)
+            for name, value in form.cleaned_data.items()
+            if name not in {'order_by', 'desc'}
+        )
+        context_data['form'] = form
+        context_data['order_by_prefix'] = query_prefix
```
ibrechin: Why is this called `order_by_prefix`? Isn't it precisely all those things that _aren't_ related to `order_by`?

ushkarev:
the fitted curves are, perhaps, not a good representation of more than a month or two into the future

PR #461, opened 2018-06-12
Send an email to users when they're locked out

ushkarev:


mtp_api/templates/mtp_auth/account_locked.html
```diff
@@ -0,0 +1,26 @@
+{% extends 'mtp_common/email_base.html' %}
+{% load i18n %}
+
+{% block inner_content %}
+  <p>
+    {% blocktrans trimmed with service_name=service_name|default:'Prisoner Money' %}
+      You’ve incorrectly entered your {{ service_name }} password too many times so your account is temporarily locked.
```
ibrechin: Should it be "Someone has attempted to log into your account with an incorrect password too many times" or something like that? I mean the theoretical point of account lockouts is to prevent someone else from brute forcing accounts, so if we say "you've" it sounds kind of weird.

PR #469, opened 2018-08-17
Split role management

ushkarev:
so that user admins can only manage people in the same role (rather than same prison OR same role)


ushkarev:
retest this please

PR #475, opened 2018-08-31
User flags

ushkarev:
NB: this now allows non-admin users editing themselves and self-deactivating

### DIFF
--- a/mtp_api/apps/mtp_auth/serializers.py
+++ b/mtp_api/apps/mtp_auth/serializers.py
@@ -9,8 +9,11 @@ from mtp_common.tasks import send_email
 from rest_framework import serializers
 from rest_framework.validators import UniqueValidator
 
-from .models import PrisonUserMapping, Role, ApplicationUserMapping, FailedLoginAttempt, AccountRequest
-from .validators import CaseInsensitiveUniqueValidator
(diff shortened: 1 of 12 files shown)


mtp_api/apps/mtp_auth/serializers.py
```diff
@@ -177,11 +196,13 @@ def update(self, user, validated_data):
         updating_user = self.context['request'].user
         was_user_admin = user.groups.filter(pk=user_admin_group.pk).exists()
 
-        if user.pk == updating_user.pk and ('user_admin' in validated_data or 'role' in validated_data):
+        disallowed_self_updates = ('user_admin', 'role', 'is_locked_out')
```
ibrechin: Possibly this should be reversed to `allowed_self_updates` - I think the list of things people should be able to update themselves is shorter than what they shouldn't.

ibrechin:
Ok I had an idea - rather than exposing flags directly via the API, use flags on the backend to store the user information, but manipulation of those flags is instead done by specific `actions` endpoints, as we have on e.g. credits for review, credit etc. That way you can restrict the ways in which flags can be edited, via both the available actions and the user permissions.

Thoughts?

PR #479, opened 2018-09-19
Use lowercase titles for our services in emails

ushkarev:
because it's the alleged GDS content style


ushkarev:
retest this please

PR #492, opened 2019-02-08
Security prison selection

ibrechin:


mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -1306,6 +1306,115 @@ def test_cannot_change_role_in_own_prison_but_different_app(self):
...
+        response = self._update_user(user, user.username, user_data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        updated_user = User.objects.get(username=user.username)
+        updated_prisons = PrisonUserMapping.objects.get_prison_set_for_user(
+            updated_user
+        )
+        self.assertEqual(len(updated_prisons), len(new_prisons))
+        self.assertEqual(
+            list(updated_prisons.values_list('nomis_id', flat=True)),
+            list(new_prisons.values_list('nomis_id', flat=True))
```
ushkarev: 🤞 the order is stable. could use `set` instead of `list`

PR #508, opened 2019-04-30
Add ECR clean-up tasks

ushkarev:
…deleting untagged images and all images in a non-master branch except the current (i.e. the latest in a PR)
…so that ECR policy limit of 40 images is not hit


ushkarev:
retest this please

PR #521, opened 2019-07-16
Simple notifications

ushkarev:
This is a bit of a complicated change :(

Because the plan for notifications radically changed during development, there is work here that's been undone. It seemed like a good idea to not drop the history of the older version of notifications for reference as there's a chance that some of it will come back in future iterations.

Everything before `Merge branch 'notifications-master' into simple-notifications` had already been reviewed and was awaiting merging, so it's only the subsequent changes that are of interest to this PR.


ushkarev:
Main changes since the previous iteration of notifications:
- removal of time-based security profile totals
- removal of most notification rules (only monitored prisoners and payment sources remain)
- notifications are now retrieved by date ranges so a new view is needed to hint at the correct date range for xx days to be returned

mtp_api/apps/credit/views.py
```diff
@@ -79,8 +79,8 @@ def get_field_filter(field):
                         return models.Q(**{'%s__startswith' % field: amount})
                 elif field == 'sender_name':
                     return (
-                        models.Q(**{'transaction__sender_name__icontains': word})
-                        | models.Q(**{'payment__cardholder_name__icontains': word})
+                        models.Q(**{'transaction__sender_name__icontains': word}) |
+                        models.Q(**{'payment__cardholder_name__icontains': word})
```
marcofucci: Actually maybe we can keep the previous version. I think people suggest having the operators at the beginning of the line so that it's easier to scan instead of scrolling to the end of the lines with different lengths.

mtp_api/apps/transaction/utils.py
```diff
@@ -1,27 +1,34 @@
...
+            text_amount = '{:0,.1f}k'.format(pounds / 1000)
+    else:
+        text_amount = '{:0,.2f}'.format(pounds)
+        if trim_empty_pence and text_amount.endswith('.00'):
+            text_amount = text_amount[:-3]
+    if pound_sign:
+        text_amount = '£' + text_amount
     if amount_pence < 0:
-        return '-' + text_amount
+        text_amount = '-' + text_amount
```
marcofucci: Non-critial: format or f-strings would be better but not a big deal here

mtp_api/apps/security/views.py
```diff
@@ -39,7 +44,40 @@ def filter(self, qs, value):
...
+
 class SenderProfileListFilter(django_filters.FilterSet):
+    sender = MultipleFieldCharFilter(
+        field_name=(
+            'bank_transfer_details__sender_name',
+            'debit_card_details__cardholder_name__name',
+            'debit_card_details__sender_email__email',
+        ),
+        lookup_expr='icontains'
+    )
```
marcofucci: Not really sure about this `sender`, it's not obvious that it searches for email address as well and it's quite similar to the `search` field I added in my PR. Also, is it actually used?

mtp_api/apps/security/views.py
```diff
@@ -104,28 +144,54 @@ class Meta:
...
     )
 
+    def get_monitor_object(self):
+        profile = self.get_object()
+        bank_details = profile.bank_transfer_details.first()
+        card_details = profile.debit_card_details.first()
+        if bank_details:
+            return bank_details.sender_bank_account
+        elif card_details:
+            return card_details
```
marcofucci: `return None` otherwise

mtp_api/apps/notification/management/commands/send_notification_emails.py
```diff
@@ -0,0 +1,153 @@
...
+from notification.models import Event, EmailNotificationPreferences
+from notification.rules import ENABLED_RULES
+from security.models import PrisonerProfile, BankAccount, DebitCardSenderDetails
+
+EMAILS_STARTED_FLAG = 'notifications-started'
+
+
+class Command(BaseCommand):
+    def handle(self, **options):
+        frequency = EMAIL_FREQUENCY.DAILY
```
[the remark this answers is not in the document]
ushkarev: they're not supported in the security tool currently. do you think it's better to:
- remove the other choices completely?
- support them as a parameter to this command even if they won't be called by scheduled commands?

mtp_api/apps/notification/management/commands/send_notification_emails.py
```diff
@@ -0,0 +1,153 @@
...
+
+from notification.constants import EMAIL_FREQUENCY, get_notification_period
+from notification.models import Event, EmailNotificationPreferences
+from notification.rules import ENABLED_RULES
+from security.models import PrisonerProfile, BankAccount, DebitCardSenderDetails
+
+EMAILS_STARTED_FLAG = 'notifications-started'
+
+
+class Command(BaseCommand):
```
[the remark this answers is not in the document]
ushkarev: which option do you prefer:
- new model just for this command with fields: user, date
- new model for this and other emails with fields: user, date, email_code, maybe even a json obj to store additional info

mtp_api/apps/credit/migrations/0032_merge_20190218_1205.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('credit', '0031_auto_20181106_1641'),
+        ('credit', '0031_private_estate'),
+    ]
+
+    operations = [
+    ]
```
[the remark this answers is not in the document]
ushkarev: it has 2 parent migrations in `credit` app and is a merge migration

mtp_api/apps/credit/migrations/0031_auto_20181106_1641.py
```diff
@@ -0,0 +1,45 @@
...
+        ),
+        migrations.AddIndex(
+            model_name='credit',
+            index=models.Index(fields=['prisoner_number', 'id'], name='credit_cred_prisone_49e443_idx'),
+        ),
+        migrations.AddIndex(
+            model_name='credit',
+            index=models.Index(fields=['-prisoner_number', 'id'], name='credit_cred_prisone_ffb7bb_idx'),
+        ),
+    ]
```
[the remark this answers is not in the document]
ushkarev: my notebook is in a crate, but i wrote down something like ~2min on my local machine for this whole batch of migrations. the previous, more complex, iteration of notifications took ~15min.

mtp_api/apps/security/views.py
ushkarev: yup, the purpose of this field was exactly the same as yours. we added it back in may, but as you can see it never got merged :) i'll delete it from this branch

ibrechin:
> * removal of time-based security profile totals

RUINED

mtp_api/apps/security/tests/test_views.py
```diff
@@ -126,16 +148,100 @@ def test_list_credits_for_sender(self):
...
+            bank_transfer_details__recipient_bank_account__monitoring_users=user
+        )
+        data = self._get_list(user, monitoring=True)['results']
+
+        self.assertEqual(len(data), recipient_profiles.count())
+        for recipient in recipient_profiles:
+            self.assertTrue(recipient.id in [d['id'] for d in data])
+
+
+class RecipientDisbursementListTestCase(SecurityViewTestCase):
```
marcofucci: I would call this `RecipientProfileDisbursementListTestCase` to make is consistent with the view name

mtp_api/apps/security/tests/test_views.py
```diff
@@ -126,16 +148,100 @@ def test_list_credits_for_sender(self):
...
+        data = self._get_list(
+            self._get_authorised_user(), path_params=[recipient.id]
+        )['results']
+        self.assertGreater(len(data), 0)
+
+        self.assertEqual(
+            len(recipient.disbursements.all()), len(data)
+        )
+        for disbursement in recipient.disbursements.all():
+            self.assertTrue(disbursement.id in [d['id'] for d in data])
```
marcofucci: I can't find any test for the `recipient-disbursements-detail` one.

mtp_api/apps/notification/management/commands/send_notification_emails.py
```diff
@@ -0,0 +1,189 @@
...
+    def handle(self, **options):
+        frequency = EMAIL_FREQUENCY.DAILY
+        period_start, period_end = get_notification_period(frequency)
+        events = get_events(period_start, period_end)
+
+        email_context = {
+            'period_start': period_start,
+            'notifications_url': (
+                f'{settings.NOMS_OPS_URL}/security/notifications/#date-{period_start.date().isoformat()}'
```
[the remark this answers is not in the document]
ushkarev: that's right. we used to have a separate view with just that one date's worth of notifications, but gavin decided to remove it.

mtp_api/apps/notification/models.py
```diff
@@ -0,0 +1,78 @@
...
+from security.models import (
+    SenderProfile, RecipientProfile, PrisonerProfile,
+)
+
+
+def validate_rule_code(value):
+    from notification.rules import RULES
+
+    if value not in RULES:
+        raise ValidationError(_('"%s" is not a recognised rule') % value)
```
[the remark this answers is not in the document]
ushkarev: it's a translated string so `value` cannot be interpreted with a format string

PR #523, opened 2019-07-22
Add simple_search search param to credit list view

marcofucci:
This adds a new `simple_search` search param to the credit list view which makes an icontains lookup on multiple text fields.
It's different from the `search` field which is more sophisticated and searches across more fields of different types.

@ushkarev I've decided not to parse the prisoner_number as I don't think the change in performance is big enough to justify the added complexity. Even if the text inputted by the user matches a prisoner number, we would still need to run ICONTAINS in OR on the other fields in case the same text appears in other places. Therefore, we would only save one ICONTAINS (which would be replaced by an IEXACT) which is minimal in my opinion.


marcofucci:
retest this please

marcofucci:
retest this please

PR #527, opened 2019-08-13
Notification report generation

ushkarev:
only includes simple rules and monitoring rules apply when _somebody_ monitors, not a given user

### DIFF
--- a/mtp_api/apps/notification/management/commands/send_notification_report.py
+++ b/mtp_api/apps/notification/management/commands/send_notification_report.py
@@ -36,11 +36,15 @@ class Command(BaseCommand):
             try:
                 validate_email(email)
             except ValidationError:
-                self.stderr.write(f'"{email}" is not valid email address')
-                return
+                raise CommandError(f'"{email}" is not valid email address')
+
+        codes = options['rules'] or RULES.keys()
+        rules = [RULES[code] for code in codes]
 
         period_start, period_end = report_period(options['since'], options['until'])
+        self.generate_reports(period_start, period_end, rules, emails)
 
+    def generate_reports(self, period_start, period_end, rules, emails):
         period_end_inclusize = period_end - datetime.timedelta(days=1)
         if period_start == period_end_inclusize:
             period_filename = period_start.date().isoformat()
@@ -68,8 +72,6 @@ class Command(BaseCommand):
         bundle_path = pathlib.Path(tempfile.mkdtemp()).absolute() / period_filename
         bundle_path.mkdir(parents=True)
 
-        codes = options['rules'] or RULES.keys()
-        rules = [RULES[code] for code in codes]
         try:
             for rule in rules:
                 if Credit in rule.applies_to_models:
@@ -162,13 +164,13 @@ def create_zip(period_filename, bundle_path):
 def send_zip(period_description, zip_file, emails):
     email = AnymailMessage(
         subject=f'Prisoner money notifications for {period_description}',
-        body=f'''
+        body=f"""
 OFFICIAL SENSITIVE
 
 Please find attached, the prisoner money notifications reports for {period_description}.
 
 If you have any queries, contact {settings.TEAM_EMAIL}.
-        '''.strip(),
+        """.strip(),
         from_email=default_from_address(),
         to=emails,
         tags=['notifications-report'],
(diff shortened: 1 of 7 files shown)


mtp_api/apps/notification/management/commands/send_notification_report.py
```diff
@@ -0,0 +1,297 @@
...
+        parser.add_argument('--rules', nargs='*', choices=RULES.keys(), help='Notification rule codes')
+        parser.add_argument('emails', nargs='+', help='Email addresses to send reports to')
+
+    def handle(self, **options):
+        emails = options['emails']
+        for email in emails:
+            try:
+                validate_email(email)
+            except ValidationError:
+                self.stderr.write(f'"{email}" is not valid email address')
```
marcofucci: Could this be replaced by raising a `CommandError` like in the other cases maybe?

mtp_api/apps/notification/management/commands/send_notification_report.py
```diff
@@ -0,0 +1,297 @@
...
+    with zipfile.ZipFile(zip_path, 'w') as z:
+        for path in report_paths:
+            z.write(path, path.name)
+    return zip_path
+
+
+def send_zip(period_description, zip_file, emails):
+    email = AnymailMessage(
+        subject=f'Prisoner money notifications for {period_description}',
+        body=f'''
```
marcofucci: flake8 wants `"""` instead

mtp_api/apps/notification/management/commands/send_notification_report.py
```diff
@@ -0,0 +1,297 @@
...
+        tags=['notifications-report'],
+    )
+    email.attach_file(str(zip_file), mimetype='application/zip')
+    email.send()
+
+
+class Serialiser:
+    def __init__(self, rule):
+        self.rule = rule
+        self.monitored_rule = isinstance(rule, MonitoredRule)
```
marcofucci: I would call this `is_monitored_rule` so that it's obvious it's a bool instead of a rule

mtp_api/apps/notification/management/commands/send_notification_report.py
```diff
@@ -0,0 +1,297 @@
...
+    def serialise(self, record):
+        row = [
+            self.rule.description.replace('you are monitoring', 'someone monitors'),
+            self.internal_id(record),
+        ]
+        if self.monitored_rule:
+            row.insert(1, self.rule.get_event_trigger(record).get_monitoring_users().count())
+        return row
+
+    def internal_id(self, record):
```
marcofucci: I would call this `get_internal_id` to make it obvious it's not a property

mtp_api/apps/notification/management/commands/send_notification_report.py
```diff
@@ -0,0 +1,297 @@
...
+        return super().get_headers() + [
+            'Date received', 'Date credited',
+            'Amount',
+            'Prisoner number', 'Prisoner name', 'Prison',
+            'Sender name', 'Payment method',
+            'Bank transfer sort code', 'Bank transfer account', 'Bank transfer roll number',
+            'Debit card number', 'Debit card expiry', 'Debit card billing address',
+            'Email', 'IP address',
+            'Status',
+            'NOMIS transaction',
```
marcofucci: One value per line would be easier to read I think, same for the `serialise` method

mtp_api/apps/notification/management/commands/send_notification_report.py
```diff
@@ -0,0 +1,297 @@
...
+    def serialise(self, record: Credit):
+        credited_at = find_log_date(record, CREDIT_LOG_ACTIONS.CREDITED)
+        return super().serialise(record) + [
+            format_csv_datetime(record.received_at), format_csv_datetime(credited_at),
+            format_amount(record.amount),
+            record.prisoner_number, record.prisoner_name, record.prison.short_name,
+        ] + self.serialise_sender(record) + [
+            CREDIT_STATUS.for_value(record.status).display,
+            record.nomis_transaction_id,
+        ]
```
marcofucci: I usually tend to use a [DictWriter](https://docs.python.org/3/library/csv.html#csv.DictWriter) to avoid the extra complexity of remembering which header goes with which row.
The serialise could create a dict per row and then you can get the first record, get its keys and you have the headers for free.

It's also easier to read instead of having to figure out what the value '' relates to.

mtp_api/apps/notification/rules.py
```diff
@@ -124,12 +122,14 @@ def get_event_trigger(self, record):
     'MONS': MonitoredRule(
         'MONS',
         'Credits for senders you are monitoring',
+        applies_to_models=(Credit,),
         profile='sender_profile',
     ),
     # disabled rules
```
marcofucci: These are not really `disabled` anymore, aren't they?

PR #528, opened 2019-08-13
Use PostcodeFilter in disbursements filtering

marcofucci:
This:
- moves PostcodeFilter to core app
- adds tests for PostcodeFilter
- replaces disbursements postcode filter field with PostcodeFilter
- refactor disbursements tests so that they test more scenarios and further scenarios can be easily added


marcofucci:
retest this please

PR #529, opened 2019-08-20
Add extra rules to notification report

ushkarev:


ushkarev:
retest this please

mtp_api/apps/notification/rules.py
```diff
@@ -114,6 +118,63 @@ def get_event_trigger(self, record):
...
+        return None
+
+    def triggered(self, record):
+        profile = self.get_event_trigger(record)
+        if not profile or profile == self.shared_profile:
+            return False
+
+        records_of_same_type = self.get_profile_records_of_same_type(profile, record)
+        field_to_count = self.kwargs['count']
+        count = records_of_same_type.values(field_to_count).distinct().count()
```
ibrechin: If only this count was kept updated and stored somewhere...

ushkarev: where do i click to report abuse

ibrechin: "this guy keeps commenting on my PRs and being right about everything"

PR #532, opened 2019-09-04
User event log ops search

marcofucci:
This 

1. adds a new django app called `user_event_log` which can be used to keep a record of specific events that occur while the user is using the system.

2. Logs search calls in the database from the intelligence tool


marcofucci:
retest this please

mtp_api/apps/user_event_log/utils.py
```diff
@@ -0,0 +1,13 @@
+from user_event_log.models import UserEvent
+
+
+def record_user_event(request, type_, data=None):
+    """Records a user event in the database."""
+    event = UserEvent(
+        user=request.user,
+        type=type_,
```
ushkarev: how do you feel about `name` or `kind` or similar to avoid the underscore?

PR #537, opened 2019-09-27
Add a simple CSV-dumping management command

ushkarev:
for exporting records to analytical platform

### DIFF
--- a/mtp_api/apps/core/management/commands/dump_for_ap.py
+++ b/mtp_api/apps/core/management/commands/dump_for_ap.py
@@ -23,36 +23,51 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         super().add_arguments(parser)
-        parser.add_argument('--since', help='Since date (inclusive)')
-        parser.add_argument('--until', help='Until date (exclusive)')
+        parser.add_argument('--after', help='Modified after date (inclusive)')
+        parser.add_argument('--before', help='Modified before date (exclusive)')
         parser.add_argument('type', choices=list(Serialiser.serialisers), help='Type of object to dump')
         parser.add_argument('path', type=argparse.FileType('wt'), help='Path to dump data to')
 
     def handle(self, *args, **options):
-        since = date_argument(options['since'])
-        until = date_argument(options['until'])
-        if since and until and until <= since:
-            raise CommandError('"--until" must be after "--since"')
+        after = date_argument(options['after'])
+        before = date_argument(options['before'])
+        if after and before and before <= after:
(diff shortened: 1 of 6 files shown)


mtp_api/apps/core/management/commands/dump_for_ap.py
```diff
@@ -0,0 +1,183 @@
...
+        'Internal ID', 'URL',
+        'Date received', 'Date credited',
+        'Amount',
+        'Prisoner number', 'Prisoner name', 'Prison',
+        'Sender name', 'Payment method',
+        'Bank transfer sort code', 'Bank transfer account', 'Bank transfer roll number',
+        'Debit card number', 'Debit card expiry', 'Debit card billing address',
+        'Sender email', 'Sender IP address',
+        'Status',
+        'NOMIS transaction',
```
marcofucci: One per line would be easier to read.

mtp_api/apps/core/management/commands/dump_for_ap.py
```diff
@@ -0,0 +1,183 @@
...
+            'Date received': record.received_at,
+            'Date credited': record.log_set.get_action_date(CREDIT_LOG_ACTIONS.CREDITED),
+            'Amount': format_amount(record.amount),
+            'Prisoner number': record.prisoner_number or 'Unknown',
+            'Prisoner name': record.prisoner_name or 'Unknown',
+            'Prison': record.prison.short_name if record.prison else 'Unknown',
+            'Status': status,
+            'NOMIS transaction': record.nomis_transaction_id,
+        }
+        row.update(self.serialise_sender(record))
```
marcofucci: You could use:
```
**self.serialise_sender(record)
```
in the same dict so that the update line is not lost.

mtp_api/apps/core/tests/test_dump_for_ap.py
```diff
@@ -0,0 +1,85 @@
...
+        with self.assertRaises(CommandError):
+            call_command('dump_for_ap')
+        with self.assertRaises(CommandError):
+            call_command('dump_for_ap', 'notifications')
+        with self.assertRaises(CommandError):
+            call_command('dump_for_ap', 'credits')
+        with self.assertRaises(CommandError):
+            call_command('dump_for_ap', 'credits', '-', after='tomorrow')
+        with self.assertRaises(CommandError):
+            call_command('dump_for_ap', 'credits', '-', before='2019-08-01', after='2019-09-01')
```
marcofucci: It would be good to say why we expect each of these tests to fail, a comment would do for example.

PR #539, opened 2019-10-08
Add new notification target for Send Money

marcofucci:
This adds a new send_money_landing target for notifications to appear on the landing page of the send money service.

I've changed the migration file directly as there's no database change so having a new file for this change is a bit of a waste.

I've only added an area for the landing page, not sure if we want one for all pages as well?


marcofucci:
Retest this please

PR #553, opened 2019-11-21
Add Check model and view to get list of check objects

marcofucci:
This:

- creates a `Check` model
- adds a serializer and an endpoint to get the list of check objects optionally filtered by status

Only users in the FIU group or with the right permissions will have access to this endpoint.


mtp_api/apps/security/urls.py
```diff
@@ -35,6 +35,9 @@
 
 router.register(r'searches', views.SavedSearchView)
 
+router.register(r'security/checks', views.CheckView, basename='security-check')
```
marcofucci: @ushkarev do you like `security/checks` or should we use `credits/checks` instead?

ushkarev: i prefer `security/checks` in case other models need to be included in the future

PR #574, opened 2020-01-15
Tighten credit frequency rules

ushkarev:
…such that more than 2 events need to happen in 1 week (previously more than 4 in 4 weeks)

### DIFF
--- a/mtp_api/apps/notification/rules.py
+++ b/mtp_api/apps/notification/rules.py
@@ -161,12 +161,12 @@ class MonitoredRule(BaseRule):
 
 
 class CountingRule(BaseRule):
-    def __init__(self, *args, profile, count, limit, **kwargs):
+    def __init__(self, *args, profile, count, limit, days, **kwargs):
         """
-        Counts a field (`count`) on records of the same type on related `profile` over the past 4 weeks.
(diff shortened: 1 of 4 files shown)


mtp_api/apps/notification/rules.py
```diff
@@ -284,12 +284,13 @@ def get_profile_records_of_same_type(self, profile, record):
...
         'CSFREQ',
-        description='More than {limit} credits from the same debit card or bank account to any prisoner in 4 weeks',
+        description='More than {limit} credits from the same debit card or bank account to any prisoner in a week',
         abbr_description='freq. source',
         applies_to_models=(Credit,),
         profile='sender_profile',
         count='pk',
-        limit=4,
+        limit=2,
+        days=8,
```
ibrechin: Are you using the dude from bodybuilding forums definition of a week?

ushkarev: no idea what you mean, but the rules were always inclusive of today and the day that is 4 weeks ago. makes sense to do the same for 1 week :D

ibrechin: you can google "bodybuilding forums days in a week" there are many articles about it and it's quite funny

PR #575, opened 2020-01-23
Upload dumps to S3 for Analytical Platform

georgeschena:
### WHAT
Implemented a way to save a days worth of transactions to S3 

### WHY
This is so the AP team can then use the data for analysis


mtp_api/apps/core/tests/test_upload_dump_for_ap.py
```diff
@@ -0,0 +1,29 @@
...
+
+class UploadDumpForAp(TestCase):
+    def test_invalid_arguments(self):
+        with self.assertRaises(CommandError, msg='Should not be missing path to file'):
+            call_command('upload_dump_for_ap')
+
+        with self.assertRaises(CommandError, msg='Should not be missing bucket name'):
+            call_command('upload_dump_for_ap', 'mtp_api/apps/core/fixtures/test_file.txt')
+
+    @mock_s3
```
ushkarev: i think `moto` isn't particularly necessary here and the test could be a lot simpler. in essence, the uploading command just passes string from the command line directly to the boto library without processing. mocking in this way involves pretending to create a bucket, upload a file and then downloading it: 2 of those steps do not happen in the management command and the third is entirely handled by boto. what i think would be better is adding a couple of lines to the other test such that `upload_to_s3` is mocked and the check is simply to see that it was called with the expected generated arguments

requirements/dev.txt
```diff
@@ -2,3 +2,5 @@
 -r base.txt
 
 money-to-prisoners-common[testing]>=9.16,<9.17
+boto3>=1.11
```
ushkarev: since the command that uses boto could be useful to run both locally and remotely, this line ought to go into base.txt: that way it's available to everyone

requirements/dev.txt
```diff
@@ -2,3 +2,5 @@
...
+moto
```
ushkarev: as above, i think moto isn't really necessary as there's no multi-step process using boto: just calling one method to upload a file can be mocked on the `upload_to_s3` function

mtp_api/apps/core/fixtures/test_file.txt
```diff
@@ -0,0 +1 @@
+is awesome
```
ushkarev: (see below first) i suspect this file won't be needed because most of `test_upload_dump_for_ap.py` will go away

mtp_api/apps/core/management/commands/dump_and_upload_for_ap.py
```diff
@@ -0,0 +1,53 @@
...
+
+        object_name = todays_date()
+
+        dump_for_ap_command(file_path_credits, file_path_disbursements)
+        upload_dump_for_ap_command(file_path_credits, file_path_disbursements, bucket_name, object_name)
+
+
+def dump_for_ap_command(file_path_credits, file_path_disbursements):
+    call_command('dump_for_ap', 'disbursements', file_path_credits, after=yesterdays_date(), before=todays_date())
+    call_command('dump_for_ap', 'credits', file_path_disbursements, after=yesterdays_date(), before=todays_date())
```
ushkarev: the date could change between these two lines, it would be safer to only calculate the ranges once
(although actually, this point is irrelevant if you have the after/before as arguments from the point above)

mtp_api/apps/core/management/commands/dump_and_upload_for_ap.py
```diff
@@ -0,0 +1,53 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument('bucket_name', help='AWS S3 global bucket name')
+
+    def handle(self, *args, **options):
+        file_path_credits = credits_temp_file_generation()
+        file_path_disbursements = disbursements_temp_file_generation()
+
+        bucket_name = options['bucket_name']
+
+        object_name = todays_date()
```
ushkarev: `object_name` is perhaps not the best name for this. it makes the `f'{object_name}_credits'` further down not really do what it's saying. maybe `date` or `date_range`?

mtp_api/apps/core/management/commands/dump_and_upload_for_ap.py
```diff
@@ -0,0 +1,53 @@
...
+def todays_date():
+    return datetime.date.today().strftime('%Y-%m-%d')
+
+
+def yesterdays_date():
+    return datetime.date.fromordinal(datetime.date.today().toordinal()-1).strftime('%Y-%m-%d')
+
+
+def credits_temp_file_generation():
+    return NamedTemporaryFile().name
```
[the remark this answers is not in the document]
ushkarev: i'd recommend a safer approach such as: [send_notification_report.py#L53](https://github.com/ministryofjustice/money-to-prisoners-api/blob/master/mtp_api/apps/notification/management/commands/send_notification_report.py#L53)
… what that does is make a temporary directory into which you put the files that you need, then once the `with` block is finished, the whole folder is deleted

mtp_api/apps/core/tests/test_dump_and_upload_for_ap.py
```diff
@@ -0,0 +1,47 @@
+from django.core.management import call_command
```
ushkarev: as above, tiny thing about ordering of imports. i'm being pedantic, but i think consistency is more important than any specific rules over any other specific rules.

mtp_api/apps/core/tests/test_dump_and_upload_for_ap.py
```diff
@@ -0,0 +1,47 @@
...
+from django.test import TestCase
+from unittest import mock
+import datetime
+
+
+class TestDumpAndUploadForAp(TestCase):
```
ushkarev: i think these tests will change a bit as a result of other changes

mtp_api/apps/core/management/commands/dump_and_upload_for_ap.py
```diff
@@ -0,0 +1,53 @@
+import textwrap
+import datetime
+from tempfile import NamedTemporaryFile
+
+from django.core.management import call_command
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
```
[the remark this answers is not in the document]
georgeschena: Spoken with Igor, we both think there is not a useful enough use case to make this change FOR NOW

mtp_api/apps/core/management/commands/dump_and_upload_for_ap.py
ushkarev: the date could still change between these lines

ushkarev: oh and you have disbursements and credits the wrong way round

mtp_api/apps/core/tests/test_dump_and_upload_for_ap.py
```diff
@@ -0,0 +1,32 @@
...
+
+        assert mock_upload_for_ap_command.called
+
+    @mock.patch('core.management.commands.upload_dump_for_ap.upload_to_s3')
+    @override_settings(ANALYTICAL_PLATFORM_BUCKET='my-bucket')
+    def test_upload_to_s3_is_called(self, mock_upload_to_s3):
+
+        call_command('dump_and_upload_for_ap')
+
+        assert mock_upload_to_s3.called
```
ushkarev: when we were talking about not needing to know what `` was called with, that was specific to it since it will always be called with effectively random strings. for `upload_to_s3` we ought to check that the it's called twice with the expected bucket name and 2 different object names. that's pretty much the effect of this feature: deciding what to put into which bucket.

mtp_api/apps/core/tests/test_upload_dump_for_ap.py
```diff
@@ -0,0 +1,35 @@
...
+            call_command('upload_dump_for_ap')
+        with self.assertRaises(CommandError, msg='bucket_name missing'):
+            call_command('upload_dump_for_ap', 'path/to/file')
+        with self.assertRaises(CommandError, msg='object_name missing'):
+            call_command('upload_dump_for_ap', 'path/to/file', 'bucket-name')
+
+
+@mock.patch('core.management.commands.upload_dump_for_ap.upload_to_s3')
+@override_settings(ANALYTICAL_PLATFORM_BUCKET='my-bucket')
+def test_upload_to_s3_is_called(self, mock_upload_to_s3):
```
ushkarev: this will never run as it's not part of a test case

PR #578, opened 2020-01-30
Performance predictions

ushkarev:


mattbee:
After taking some time to find the point at which these predictive curves were being used, some more detail in b12e97d might have been useful - but that's a minor point, feel free to skip.

PR #581, opened 2020-02-04
Add cron job for dump_and_upload_for_ap

georgeschena:
### WHAT
Add Cron job for dump_and_upload_for_ap at 00:10 everyday

### WHY
So we can send data to S3 daily


ushkarev:
actually, i coming back to this, i forgot to think about concurrency. this will make all 3 pods compile and upload the file to AP. it would probably work but create thrice the load on the db and all other parts involved.

i think we ought to not have this triggered by uwsgi's cron feature. there's nothing in the management command to prevent concurrent execution (here's an [example](https://github.com/ministryofjustice/money-to-prisoners-api/blob/b1562b5e9741a1203beed2ab810be2d26949cef6/mtp_api/apps/core/management/commands/clean_up.py#L21)). i think instead, we should just register this in the django admin interface.

ushkarev:
this was registered as a scheduled task in django admin, so cron task is not needed

PR #583, opened 2020-02-04
Refactor file handling in Analytical Platform dump

mattbee:
In working with the `dump_for_ap` and uplaoding the results of the
export command, we ran into problems with the last 8kb of the exported
data remaining in the buffer when the file was worked with in the
command itself.

Although the final implementation will not be called while any export
data remains in the buffer, this was an unexpected complexity that could
be avoided.

Given that a `with` block closes a file when its finished, we think this
simplifies the export itself. Removing the `argsparse.FileType` however
meant that we didn't have ahook to stdout, so the tests had to update
to use temporary files.

This should also make the conversion to a JSON Lines format easier, as
we need to work with the data export without `csv.DictWriter` help.

### DIFF
--- a/mtp_api/apps/core/management/commands/dump_for_ap.py
+++ b/mtp_api/apps/core/management/commands/dump_for_ap.py
@@ -1,4 +1,3 @@
-import argparse
 import csv
 import datetime
 import textwrap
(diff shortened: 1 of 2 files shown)


mattbee:
> there's a slight logic change in the tests: the credit ids read from the csv aren't sorted, whereas what they're being compared to _is_ still sorted. it seems that at the moment the data generation already does ids in the same order, so 🤷‍♀ (just though i'd point it out)

Following our chat, since the AP dump orders by primary key, this is all OK and tests will still be reliable.

mattbee:
Can we restructure the commits on this branch a bit before we merge. I'll have a look if I run out of things today, but just want to put this here before we merge.

PR #584, opened 2020-02-06
Convert output file to JSON Lines format

mattbee:
This change is converting the previous data format of CSV to JSON Lines[1] as requested by the data team.

The change simplifies the code a little bit following the previous refactor (#583).

### Questions
- Should we worry about any previous data breaking the JSON encoding?
- Will all systems that we are exporting to use the right EOL sequence (i.e. will `\n` work in all environments this code will run on?)

1. http://jsonlines.org/

### DIFF
--- a/mtp_api/apps/core/management/commands/dump_for_ap.py
+++ b/mtp_api/apps/core/management/commands/dump_for_ap.py
@@ -1,6 +1,6 @@
-import csv
 import datetime
 import textwrap
+import json
 
 from django.conf import settings
 from django.core.management import BaseCommand, CommandError
@@ -36,18 +36,17 @@ class Command(BaseCommand):
         record_type = options['type']
         serialiser: Serialiser = Serialiser.serialisers[record_type]()
 
-        with open(options['path'], 'wt') as csv_file:
-            writer = csv.DictWriter(csv_file, serialiser.headers)
-            writer.writeheader()
+        with open(options['path'], 'wt') as jsonl_file:
             records = serialiser.get_modified_records(after, before)
+
             for record in records:
-                writer.writerow(serialiser.serialise(record))
+                jsonl_file.write(json.dumps(serialiser.serialise(record), default=str, ensure_ascii=False))
+                jsonl_file.write('\n')
 
 
 class Serialiser:
     serialisers = {}
     serialised_model = None
-    headers = []
 
     def __init_subclass__(cls, serialised_model):
         record_type = str(serialised_model._meta.verbose_name_plural)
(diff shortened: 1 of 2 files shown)


mtp_api/apps/core/tests/test_dump_for_ap.py
```diff
@@ -48,20 +50,24 @@ def test_valid_arguments(self, mocked_get_modified_records):
         self.assertEqual(before.date(), datetime.date(2019, 9, 1))
 
     def test_empty_results(self):
-        with captured_stdout() as stdout:
-            call_command('dump_for_ap', 'credits', '-')
-        stdout = stdout.getvalue().strip()
-        lines = stdout.splitlines()
-        self.assertEqual(len(lines), 1)
+        filename = tempfile.NamedTemporaryFile().name
```
georgeschena: Maybe use `tempfile.NamedDirectory`?

ushkarev:
my thoughts on the questions:
- i wouldn't worry – there is one app using this data in AP and the airflow pipeline (which is to-be-created) would be the only thing that processes this dump
- json line format is compatible with both `\n` and `\r\n` and the airflow pipeline will be the only thing processing it so it's the only thing that needs to care. plus, pretty much no tools these days care about EOL types

ushkarev:
`Serialiser.headers` should also be removed as it's no longer used

mtp_api/apps/core/tests/test_dump_for_ap.py
ushkarev: possibly. or at least it should be consistent with the tests below such that the temporary file remains in existence rather than relying on GC kicking in late and thereby actually deleting the file produces by the next line

georgeschena: Yeah this makes sense

mattbee: Refactored this to be consistent with the other uses, it does make more sense now.

mtp_api/apps/core/management/commands/dump_for_ap.py
```diff
@@ -36,17 +35,18 @@ def handle(self, *args, **options):
...
-        writer.writeheader()
-        records = serialiser.get_modified_records(after, before)
-        for record in records:
-            writer.writerow(serialiser.serialise(record))
+
+        with open(options['path'], 'wt') as jsonl_file:
+            records = serialiser.get_modified_records(after, before)
+
+            for record in records:
+                jsonl_file.write(json.dumps(serialiser.serialise(record), default=str, ensure_ascii=False))
```
ushkarev: can you remind me what the `default=str` is for? is it the datetime type? to my mind, it's the serialisers' job to ensure that the output is in the correct format. will `default=str` here output the datetime in a timezone-aware machine readable format?

mattbee: That's right - a datetime value breaks the JSON output, and the default=str handles this.

There's likely a need for the analysis to parse the datetime from a string but I'd assume that this was a known in requesting JSON Lines format.

ushkarev: so is the output timezone aware?

ushkarev: ok i checked, and looks like it comes out _nearly_ in ISO format. i guess it's up to you what you prefer – while it's not the super standard format, it does have timezone info

PR #586, opened 2020-02-13
Add job information end point

georgeschena:
General code comments would be really useful. We don't have any specific questions, but would appreciate feedback.

Do you think the test coverage is sufficient?  And again, improvement suggestions would be really appreciated.


mtp_api/apps/mtp_auth/admin.py
```diff
@@ -34,6 +37,17 @@ class RoleAdmin(ModelAdmin):
     def all_groups(self, instance):
         return ', '.join(sorted(group.name for group in instance.groups))
 
+@admin.register(JobInformation)
+class JobInformationAdmin(ModelAdmin):
+    list_display = ('title', 'prison_estate', 'tasks', 'fullname', 'email', 'user_id')
+    list_filter = ('title', 'prison_estate')
```
ushkarev: i'm not sure how well `prison_estate` will work here since it's a free text field as far as this app is concerned

mtp_api/apps/mtp_auth/admin.py
```diff
@@ -34,6 +37,17 @@ class RoleAdmin(ModelAdmin):
     def all_groups(self, instance):
         return ', '.join(sorted(group.name for group in instance.groups))
 
+@admin.register(JobInformation)
+class JobInformationAdmin(ModelAdmin):
+    list_display = ('title', 'prison_estate', 'tasks', 'fullname', 'email', 'user_id')
```
ushkarev: i don't think `user_id` is useful to anybody, but i suppose you could have `username` instead

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -52,6 +53,61 @@ def prison_set(user):
...
 
 
+class JobInformationTestCase(APITestCase, AuthTestCaseMixin):
+    fixtures = [
+        'initial_groups.json',
+        'initial_types.json',
+        'test_prisons.json',
+    ]
+
+    url = reverse('job-information-list')
```
ushkarev: if you're going to have this at class-level, you ought to use the `reverse_lazy` form, because urls can be dynamic. it doesn't make any difference in this case, but would be good practice

ushkarev: `reverse_lazy` only looks up the url when the value is _accessed_ not at the moment it's called

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -52,6 +53,61 @@ def prison_set(user):
...
+        'title': 'Warden',
+        'prison_estate': 'Private',
+        'tasks': 'Run the show',
+    }
+
+    def setUp(self):
+        super().setUp()
+        test_users = make_test_users()
+        self.security_staff = test_users['security_staff'][0]
+        self.security_staff.current_password = 'foobar'
```
ushkarev: what is this line for? all the test users have their username as their password

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -52,6 +53,61 @@ def prison_set(user):
...
+        response = self.client.post(
+            self.url,
+            data=self.payload,
+            HTTP_AUTHORIZATION=self.get_http_authorization_for_user(self.security_staff)
+        )
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(response.data, self.payload)
+
+    def test_will_not_create_with_missing_params(self):
+        self.payload['prison_estate'] = ''
```
ushkarev: you're modifying a class-level attribute so all tests that run after this moment are affected! it's safer to just have a different dictionary defined here.

mtp_api/apps/mtp_auth/models.py
```diff
@@ -88,6 +88,13 @@ def assign_to_user(self, user):
             user.groups.add(group)
 
 
+class JobInformation(models.Model):
```
ushkarev: one thing i'd suggest, but it's no big deal, is subclassing TimeStampedModel. that way, you know the date/time when someone entered their job title. i think that'd be helpful. if you do this, you can then add 'modified' to the admin `list_display` and add an `ordering`.

mtp_api/apps/mtp_auth/admin.py
mattbee: Ah - in the form we have three options so that should control options initially. There's no other option on the question so hopefully should be OK initially.

ushkarev: fair enough

mattbee:
FYI Builds failing as NPM is down :(

mtp_api/apps/mtp_auth/models.py
mattbee: Great - I think the ordering is most useful so that's been added.

mtp_api/apps/mtp_auth/tests/test_views.py
mattbee: Good shout, have updated that to use a fresh dictionary.

mtp_api/apps/mtp_auth/admin.py
mattbee: Good point, we have updated that

mtp_api/apps/mtp_auth/tests/test_views.py
mattbee: Ah, I didn't know that - but this was also legacy code from a previous method of auth anyway - by using the post request with `HTTP_AUTHORIZATION` we don't need it anyway.

mtp_api/apps/mtp_auth/views.py
```diff
@@ -59,6 +59,10 @@ def get_queryset(self):
         return queryset
 
 
+class JobInformationViewSet(viewsets.mixins.CreateModelMixin, viewsets.GenericViewSet):
+    serializer_class = JobInformationSerializer
```
ushkarev: might be worth putting in an explicit `permission_classes = (IsAuthenticated,)` in here even though it's on by default cos the serialiser requires a proper user object

PR #587, opened 2020-02-17
WIP Save a flag when a job information record is saved

mattbee:
This adds the flag which will be used in the login process for the Intelligence app.

### Questions

- Is the create method in the Serializer the best place to save the flag record?
- We thought we might be able to access a user flags in the test (`self.security_staff.flags`) but this just returns `None` - should this be possible?

### DIFF
--- a/mtp_api/apps/mtp_auth/serializers.py
+++ b/mtp_api/apps/mtp_auth/serializers.py
@@ -43,6 +43,7 @@ class JobInformationSerializer(serializers.ModelSerializer):
 
     def create(self, validated_data):
         user = self.context['request'].user
+        user.flags.create(name='provided-job-information')
         validated_data['user'] = user
         return super().create(validated_data)
 
--- a/mtp_api/apps/mtp_auth/tests/test_views.py
+++ b/mtp_api/apps/mtp_auth/tests/test_views.py
@@ -110,6 +110,16 @@ class JobInformationTestCase(APITestCase, AuthTestCaseMixin):
         self.assertEqual(len(records), 1)
         self.assertEqual(records[0].user_id, self.security_staff.id)
 
+    def test_flag_created_when_job_info_saved(self):
+        self.client.post(
+            self.url,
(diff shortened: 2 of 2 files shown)


mtp_api/apps/mtp_auth/serializers.py
```diff
@@ -43,6 +43,7 @@ class Meta:
 
     def create(self, validated_data):
         user = self.context['request'].user
+        user.flags.create(name='provided-job-information')
```
ushkarev: yup, i'm cool with this happening here. might be nice to have the flag in `mtp_auth.constants` to be able to refer to it consistently

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -110,6 +110,16 @@ def test_user_id_added_to_job_information_record(self):
...
         self.assertEqual(records[0].user_id, self.security_staff.id)
 
+    def test_flag_created_when_job_info_saved(self):
+        self.client.post(
+            self.url,
+            data=self.payload,
+            HTTP_AUTHORIZATION=self.get_http_authorization_for_user(self.security_staff)
+        )
+
+        flag_exists = Flag.objects.filter(user_id=self.security_staff.id, name='provided-job-information').exists()
```
ushkarev: not sure i understand the second question in the PR: what this is doing makes sense to me.

could be a tiny bit slightly simpler with `…user=self.security_staff…`

mtp_api/apps/mtp_auth/serializers.py
mattbee: Another thought this morning, this does mean we have the flag declared in 2 different apps (a constant here and a constant in the __init__.py of noms-ops app).

Is it actually less fragile to update the flag from noms-ops and just have the constant set in the one app?

ushkarev: not sure it makes any real difference… your call :)

PR #588, opened 2020-03-06
Allow API to return 'decision_reason' and 'actioned_by_name'

georgeschena:


mtp_api/apps/security/serializers.py
```diff
@@ -219,19 +219,32 @@ def update(self, instance, validated_data):
...
             'rules',
             'actioned_at',
             'actioned_by',
+            'rejection_reason',
+            'full_name',
         )
 
+    def get_full_name_from_user(self, check):
+        if check.actioned_by is not None:
+            first_name = check.actioned_by.first_name
```
ushkarev: the user object already has a `get_full_name` method. also, i think it might be better to say whose full name this is. maybe `actioned_by_name` or something?

mtp_api/apps/security/serializers.py
```diff
@@ -219,19 +219,32 @@ def update(self, instance, validated_data):
...
 class CheckSerializer(serializers.ModelSerializer):
     credit = SecurityCreditSerializer()
+    full_name = serializers.SerializerMethodField('get_full_name_from_user')
 
     class Meta:
         model = Check
         fields = (
             'id',
             'credit',
+            'created',
```
ushkarev: what are you thinking of using this for? it's not quite the same as a credit's `started_at`

mtp_api/apps/security/views.py
```diff
@@ -497,6 +497,9 @@ class CheckListFilter(django_filters.FilterSet):
     credit_resolution = django_filters.CharFilter(
         field_name='credit__resolution', lookup_expr='exact',
     )
+    actioned_by = django_filters.BooleanFilter(
+        field_name='actioned_by', lookup_expr='isnull', exclude=True,
```
ushkarev: how come you need the `exclude=True`? would you not just post {actioned_by=false} to filter for anything that has a non-null actioned_by field?

mtp_api/apps/security/tests/test_views.py
```diff
@@ -1103,6 +1105,31 @@ def test_get(self):
         actual_check_data = response.json()
         self.assertCheckEqual(check, actual_check_data)
 
+    def test_get_for_an_actioned_check(self):
+        """
+        Test that the get object endpoint returns check details.
+        """
+        check = Check.objects.filter(actioned_by__isnull=False).first()
+
+        first_name = self.security_fiu_users[0].first_name
```
ushkarev: `get_full_name` is a method on the user object you can use for this already

mtp_api/apps/security/tests/test_views.py
```diff
@@ -1103,6 +1105,31 @@ def test_get(self):
...
+            ),
+            format='json',
+            HTTP_AUTHORIZATION=auth,
+        )
+
+        self.assertEqual(response.status_code, http_status.HTTP_200_OK)
+
+        actual_check_data = response.json()
+        self.assertCheckEqual(check, actual_check_data)
+        self.assertEqual(actual_check_data['full_name'], f'{first_name} {last_name}')
```
ushkarev: `assertCheckEqual` already checks for full name being present, no?

mtp_api/apps/security/views.py
ushkarev: also, could be worth having a test that uses this filter

georgeschena: I feel that this exposes the complexity to the the noms_ops app as it makes it confusing to read `params['actioned_by'] = False` rather than `params['actioned_by'] = True`. By doing `params['actioned_by'] = True` makes it very explicit that you are filtering for checks with ab `actioned_by `. This way just hides the complexity in the API rather than the front end.

mattbee: Yeah, it was a toss up between clean code and an easier to understand API - filtering by not null isn't directly possible, so we thought hiding the complexity in the API was better than having `actioned_by = False` return rows where actioned_by is a truthy value. 

https://django-filter.readthedocs.io/en/master/guide/tips.html#filtering-by-null-values

Unless you have a different solution, happy to try it if you think there's a less complex way that makes sense in the API calls too.

ushkarev: filtering by "not null" would be possible if you add it to the fields bit (similar to [here](https://github.com/ministryofjustice/money-to-prisoners-api/blob/master/mtp_api/apps/credit/views.py#L221)) but your way of doing it is fine. i agree that it's a toss up between the client side being clear (actioned_by=True means that somebody actioned) or the api being clear ("exclude nulls if filter is set to true").

mtp_api/apps/security/tests/test_views.py
```diff
@@ -1076,18 +1078,63 @@ def test_check_filtering_by_credit_resolution(self):
...
+            format='json',
+            HTTP_AUTHORIZATION=auth,
+        )
+
+        self.assertEqual(response.status_code, http_status.HTTP_200_OK)
+
+        response_data = response.json()
+
+        self.assertEqual(
+            response_data['count'],
```
ushkarev: instead of just checking you have the right number, i think you should look at the checks returned and see that they were filtered as expected

mtp_api/apps/security/tests/test_views.py
```diff
@@ -1076,18 +1078,63 @@ def test_check_filtering_by_credit_resolution(self):
...
     """
     Tests related to getting a single security check.
     """
 
-    def test_get(self):
+    def test_get_for_a_non_actioned_check(self):
         """
         Test that the get object endpoint returns check details.
         """
         check = Check.objects.first()
```
ushkarev: how do you know this is "non actioned"?

ushkarev: actually, what's the difference between `test_get_for_a_non_actioned_check` and `test_get_for_an_actioned_check`?

georgeschena: test_get_for_a_non_actioned_check = actioned_by is null
test_get_for_an_actioned_check = actioned_by is not null

georgeschena: I have changed this line to `check = Check.objects.filter(actioned_by__isnull=True).first()`

ushkarev: but this endpoint doesn't do any filtering

georgeschena: This now makes it more explicit on what the test means imo

ushkarev: no harm i guess, but i can't see what difference in logic you're testing

georgeschena: Oh I see what you're saying 🥇

georgeschena: My misunderstanding

mtp_api/apps/security/tests/test_views.py
georgeschena: I feel this test does this for us anyway? As we are asserting that we only want a response with checks which have an `'actioned_by': True` which means only checks which have an actioned_by id. Then we are checking the DB to only grab the checks which have an `actioned_by` so when asserting this should be enough? Because the query to the DB will only bring back the checks which have an actioned_by which is not null. I tried to keep the tests similar to the ones above for consistency purposes also.

georgeschena: From this test I feel this gives me enough confidence it is filtering the way we expect

ushkarev: >Then we are checking the DB to only grab the checks which have an actioned_by
you are not. you're only checking that it's the same _number_

ushkarev: the point of this test is to _ensure_ that the filter `{'actioned_by': true}` does the right thing, which is returning only non-nulls. this test only proves you get the right number back. but what you want is for it to make sure the contract is maintained if any further changes happen on the view

georgeschena: Okay this makes more sense. Ill implement this

mtp_api/apps/security/tests/test_views.py
```diff
@@ -1103,6 +1103,9 @@ def test_check_filtering_by_actioned_by(self):
             Check.objects.filter(actioned_by__isnull=False).count(),
         )
 
+        for item in response_data['results']:
+            self.assertIsNotNone(item['actioned_by'])
```
ushkarev: exactly this!

PR #591, opened 2020-04-07
Stop a test leaving temporary files

ushkarev:


ushkarev:
doh. that’s in my global gitignore file rather in our repos. i find it helpful to have ignored folders and files to avoid commit stuff that shouldn't be. shall we put `*~` into all .gitignore files?

PR #593, opened 2020-04-28
Improve reports in Django admin

ushkarev:
Reports are now consistent so that there are 2 kinds of reports for each of:

- credits
- disbursements
- digital take-up

…with the 2 kinds being:

- "report" which looks at history sliced into months/quarters/years/financial years
- "per-prison" which instead looks at prisons for a specified slice of time

Digital take-up report excludes the private estate by default since postal credits are not know there. It had previously been assumed that digital take-up is the same in the private estate.

There are now notes in all reports explaining how they work more clearly and indicating where data comes from.


mtp_api/apps/performance/prediction.py
```diff
@@ -68,49 +77,73 @@ def get_value(self, x):
...
 known_curves = {
-    'accurate_credits_by_mtp': {
+    'accurate_credits_by_mtp_with_private_estate': {
+        'curve': Logarithmic,
+        'default_params': np.array([83089128.91457231, 1.6418543562686948e-05, 1.000494203778578], dtype='float64')
+    },
+    'accurate_credits_by_mtp_without_private_estate': {
         'curve': Logarithmic,
-        'defaults': np.array([109004.08604805294, 0.022066612129535555, 1.4388566286122846], dtype='float64')
+        'default_params': np.array([78402589.76473396, 1.656958576241639e-05, 1.0005337542137802], dtype='float64')
```
Staberinde, line 86: Any reason why you're not just setting the default coefficients to unity since you're running the optimisation each time. It feels like you're increasing the risk of finding local minima if you're priming with previous coefficients?

mtp_api/templates/admin/performance/digitaltakeup/report.html
```diff
@@ -0,0 +1,225 @@
...
+        cols: [
+          {type: 'string', label: '{% trans 'Periods' %}'},
+          {type: 'number', label: '{% trans 'NOMIS digital credits' %}'},
+          {type: 'number', label: '{% trans 'NOMIS postal credits' %}'},
+          {type: 'number', label: '{% trans 'Digital credits' %}'},
+          {type: 'boolean', role: 'certainty'},
+          {type: 'number', label: '{% trans 'Scaled postal credits' %}'},
+          {type: 'boolean', role: 'certainty'}
+        ],
+        rows: {{ chart_rows_json|safe }}
```
Staberinde, line 36: As a general rule I'm not a fan of templating javascript, as it makes it difficult to extract the javascript out to a separate file. I'm not suggesting changing it now, but it might be worth looking at refactoring at some point.

mtp_api/urls.py
```diff
@@ -47,4 +43,6 @@
         _('Sorry, something went wrong'),
         content_type='text/plain', status=500,
     )),
+
+    url(r'^$', lambda request: HttpResponse(content_type='text/plain', status=204)),
```
Staberinde, line 47: Any reason for moving this down into a separate block?

mtp_api/apps/core/templatetags/simple_admin.py
```diff
@@ -20,3 +20,24 @@ def admin_choice_list(form, bound_field):
         'field': bound_field,
         'choices': bound_field.field.choices,
     }
+
+
+@register.inclusion_tag('admin/sortable-table-header.html')
+def sortable_table_header(form, ordering, title):
```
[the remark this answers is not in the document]
ushkarev: you mean how the query string is put together? no major reason. the stuff in `django.admin` does not appear reusable and i think "?o=1" is not so readable and has a more complex functionality. this way looks more like the way django rest framework + django filters do it: "?ordering=count"

mtp_api/apps/core/forms.py
```diff
@@ -44,63 +47,145 @@ class Media:
...
+        if 'ordering' in self.fields:
+            ordering_field = self.fields['ordering']
+            reversed_choices = [
+                (f'-{value}', label)
+                for value, label in ordering_field.choices
+            ]
+            ordering_field.choices.extend(reversed_choices)
+
+    @cached_property
+    def query_string_without_ordering(self):
```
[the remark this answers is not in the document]
ushkarev: this is to make table header cell links for sorting a table by that column. this form property gives you the base query string to which various ordering options are appended depending on the column

mtp_api/apps/credit/views_admin.py
```diff
@@ -0,0 +1,123 @@
...
+    template_name = 'admin/credit/credit/prison-report.html'
+    form_class = PrisonCreditReportForm
+
+    def get_context_data(self, **kwargs):
+        context_data = super().get_context_data(**kwargs)
+        form = context_data['form']
+
+        prisons = Prison.objects \
+            .exclude(nomis_id__in=self.excluded_nomis_ids) \
+            .order_by() \
```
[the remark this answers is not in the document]
ushkarev: it removes any default ordering. here it doesn't matter since the results are not used in any order. but in queries resulting in `GROUP BY` clauses, it makes a huge difference

mtp_api/apps/performance/models.py
```diff
@@ -1,13 +1,76 @@
...
+            included_prisons_sql = ''
+            credit_count_join_sql = ''
+            average_takeup_join_sql = ''
+
+        sql = f"""
+        WITH
+            {included_prisons_sql}
+            credit_count AS (
+                SELECT
+                    date_trunc('month', received_at AT TIME ZONE '{settings.TIME_ZONE}')::timestamp WITH TIME ZONE
```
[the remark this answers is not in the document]
ushkarev: that sounds interesting… could be a good thing to try. i worry that it might affect loads of different parts of the code though. also, TIME_ZONE wouldn't be anything other than Europe/London

mtp_api/apps/performance/models.py
```diff
@@ -1,13 +1,76 @@
...
+            """
+        else:
+            included_prisons_sql = ''
+            credit_count_join_sql = ''
+            average_takeup_join_sql = ''
+
+        sql = f"""
+        WITH
+            {included_prisons_sql}
+            credit_count AS (
```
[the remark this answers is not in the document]
ushkarev: i dunno if one could do it with django ORM. i don't know if it would be more readable either

mtp_api/templates/admin/performance/digitaltakeup/report.html
ushkarev: fair point :)

mtp_api/apps/performance/prediction.py
ushkarev: as it happens, the optimisation is never run server-side so this is the only place prediction curve values exist

mtp_api/urls.py
ushkarev: it seemed more like an error page / fallback than important app code. it fits less well with the group of includes at the top

mtp_api/apps/core/forms.py
```diff
@@ -128,12 +215,78 @@ def format_date(d):
...
+        if self.cleaned_data['period'] == 'quarterly':
+            yield from self.group_months_into_quarters(monthly_rows)
+        elif self.cleaned_data['period'] == 'yearly':
+            yield from self.group_months_into_years(12, monthly_rows)
+        elif self.cleaned_data['period'] == 'financial':
+            yield from self.group_months_into_years(3, monthly_rows)
+        else:
+            yield from monthly_rows
+
+    def group_months_into_quarters(self, monthly_rows):
```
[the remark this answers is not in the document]
ushkarev: i'm fairly ambivalent about `doctest`s, but not against them. i sort of like the simplicity of standard `unittest`s since they're generally in one place rather than interspersed in the code so are easy to find

mtp_api/apps/credit/views_admin.py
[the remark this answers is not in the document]
ushkarev: https://docs.djangoproject.com/en/2.0/topics/db/aggregation/#interaction-with-default-ordering-or-order-by

mtp_api/apps/core/forms.py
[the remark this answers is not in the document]
ushkarev: i doubt they'd play well with and DB functionality so would need to be carefully tailored, right?

mtp_api/apps/performance/views.py
```diff
@@ -172,13 +172,19 @@ class DigitalTakeupReport(AdminViewMixin, TemplateView):
...
+        self.exclude_private_estate = True
+
     def get_context_data(self, **kwargs):
         context_data = super().get_context_data(**kwargs)
         form = DigitalTakeupReportForm(data=self.request.GET.dict())
         if not form.is_valid():
             form = DigitalTakeupReportForm(data={})
             assert form.is_valid(), 'Empty form should be valid'
 
+        self.exclude_private_estate = form.cleaned_data['private_estate'] == 'exclude'
```
mattbee, line 174: Given `self.exclude_private_estate` is already set to true, I think it might be more readable to explicitly show that we are just changing this to False if the private estate value is not exclude:
```
if form.cleaned_data['private_estate'] != 'exclude':
    self.exclude_private_estate = False
```

Super minor thing but took me a second longer to read it in the current form.

mtp_api/templates/performance/digital-takeup-report.html
```diff
@@ -9,16 +9,13 @@
   {{ block.super }}
 
   <link rel="stylesheet" type="text/css" href="{% static 'admin/css/changelists.css' %}" />
+  <link rel="stylesheet" type="text/css" href="{% static 'stylesheets/digital-takeup-reports.css' %}"/>
   <style>
     #takeup-chart {
```
mattbee: Is it worth the other css rules to the stylesheet?
Also- I couldn't see the file linked by `{% static 'admin/css/changelists.css' %}` anywhere, where does this get pulled in from?

mtp_api/apps/core/views.py
```diff
@@ -70,6 +70,9 @@ class AdminReportView(AdminViewMixin, FormView, metaclass=MediaDefiningClass):
     """
     form_class = NotImplemented
 
+    # test prisons that should not feature in any report
+    excluded_nomis_ids = {'ZCH'}
```
mattbee, line 75: I've seen this value 'ZCH' pop up a few times in other files, can it be extracted to a global setting (I see that's pretty much what you have done for this file...) for excluded NOMIS Ids or will it never change? Or is it a test prison as doesn't exist and just exists for testing nomis?

mtp_api/apps/mtp_auth/models.py
```diff
@@ -94,6 +94,9 @@ class JobInformation(TimeStampedModel):
     prison_estate = models.CharField(max_length=255)
     tasks = models.TextField()
 
+    class Meta:
+        verbose_name_plural = 'job information'
```
mattbee, line 98: Thanks :)

mtp_api/apps/core/views.py
mattbee: Having looked, it's only in 3 places which doesn't make it essential for me yet.

mtp_api/apps/performance/views.py
ushkarev: ah the reason why it's pre-set to true is so that the attribute always exists. so doing `self. exclude_private_estate` will not raise an error. static analysis would complain if an attribute is set on self for the first time outside of the __init__ method. happy to go with you suggestion tho.

mtp_api/templates/performance/dashboard.html
```diff
@@ -12,208 +12,210 @@
...
   </script>
   <script type="text/javascript" src="{% static 'javascripts/performance-dashboard.js' %}"></script>
 </head>
-  <body>
-     <div id="disbursements_page" class="size-of-page" style="display:none">
-      <div class="main-title">
-        <h1><strong>Disbursements</strong> (Money sent by prisoners)</h1>
-      </div>
+<body>
+  <div id="disbursements_page" class="size-of-page" style="display:none">
```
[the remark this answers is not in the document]
ushkarev: oh i agree. this is not an ideal template. it's missing translations too and some ids and class names aren't that clear.

the only important change i made was removing a random `</tr>` on line 109

mtp_api/templates/performance/digital-takeup-report.html
ushkarev: that style sheet come from Django itself. it's one of the many that are automatically used in various parts of the admin interface.

mtp_api/apps/core/views.py
ushkarev: i think you're right. it would probably be a good idea to centralise this somehow. maybe on the Prison queryset/manager. it is in fact a test prison that exists in our production data, but does not in fact exist in NOMIS. in some situations we _might_ want to see it so it's not excluded super-globally.

PR #594, opened 2020-05-04
MTP-1352 Review pending credit/matching credits

Staberinde:
Ticket: https://dsdmoj.atlassian.net/browse/MTP-1352

Changes:
* Associates `prisoner_profile` and `sender_profile` on `Credit` creation
* New field on Credit, `is_counted_in_sender_profile_total` to emulate previous filtering functionality within `update_security_profiles` command
* Removal of any logic predicated on population of `credit.sender_profile`/`credit.prisoner_profile` and replacement with suitable alternative
* Modified `update_security_profiles` job so that it updates prisoner/sender aggregations exactly once per profile instead of once per credit
* Added `exclude_credit__in` to API, to facilitate omission of credit under review for queries from new noms_ops views (see https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/477)
* Fix to allow circleCI to operate on feature branches containing capital letters 
* Incremented `psycopg2-binary` to allow it to be built by python3.8
* Assorted `.gitignore`/`.dockerignore` additions


mtp_api/apps/notification/tests/test_commands.py
```diff
@@ -46,8 +50,11 @@ def setUp(self):
         test_users = make_test_users()
         self.security_staff = test_users['security_staff']
         load_random_prisoner_locations()
-        generate_payments(payment_batch=20, days_of_history=2)
-        generate_disbursements(disbursement_batch=20, days_of_history=2)
+        generate_payments(
```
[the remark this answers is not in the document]
Staberinde: This was needed to get the tests to consistently reflect the state of the code (i.e. not fail erroneously) before I added `override`in, but we can probably revert it back to 2 days now

mtp_api/apps/security/management/commands/update_security_profiles.py
```diff
@@ -128,8 +179,10 @@ def process_disbursement_batch(self, new_disbursements):
         return len(new_disbursements)
 
     def create_or_update_profiles_for_credit(self, credit):
-        sender_profile = SenderProfile.objects.create_or_update_for_credit(credit)
-        if credit.prison:
+        if not credit.sender_profile:
+            # TODO this method does not need to return, pull sender_proffile off credit object
+            sender_profile = SenderProfile.objects.create_or_update_for_credit(credit)
+        if credit.prison and credit.sender_profile and not credit.prisoner_profile:
```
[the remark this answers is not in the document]
Staberinde: Hmm, thought i'd fixed that, apologies

mtp_api/apps/security/management/commands/update_security_profiles.py
```diff
@@ -45,64 +46,114 @@ def handle_update(self, batch_size, recreate):
...
             if credit.prisoner_profile:
                 prisoner_profiles.append(credit.prisoner_profile.pk)
+            else:
+                logger.warning('Prisoner profile could not be found for credit %s', credit)
 
-        SenderProfile.objects.filter(
-            pk__in=sender_profiles
-        ).recalculate_credit_totals()
-        PrisonerProfile.objects.filter(
+        # TODO this was previous behaviour, but I feel like it might not be appropriate now
```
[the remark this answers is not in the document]
Staberinde: The legacy mode only populates sender/prisoner profiles, `calculate_credit_totals_for_sender_profiles` will still be called even for legacy credits.

mtp_api/apps/security/managers.py
```diff
@@ -253,24 +253,43 @@ def recalculate_totals(self):
...
             ), 0),
         )
+        new_credits = Credit.objects.filter(
+            prisoner_profile__in=self,
+            resolution=CREDIT_RESOLUTION.CREDITED,
+            is_counted_in_prisoner_profile_total=False
+        )
+        new_credits_ids = [c.id for c in new_credits]
+        new_credits.update(is_counted_in_prisoner_profile_total=True)
+        # This is kinda grim, but I don't like the idea of passing around stale
```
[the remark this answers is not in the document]
Staberinde: I don't think there's an explicit action here, the choice is either requery or pass around stale data (which is dangerous), this is meant more as an explanation as to the design decisions taken here.

mtp_api/apps/credit/managers.py
```diff
@@ -6,6 +8,9 @@
 from credit.constants import LOG_ACTIONS, CREDIT_STATUS, CREDIT_RESOLUTION
 
 
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
Staberinde: The issue with having every file log under the same namespace is that it can be difficult to trace back to where the logging statement was emitted. Is there a reason for having all files use the same logging namespace?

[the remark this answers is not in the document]
ushkarev: the logging settings explicitly set how the 'mtp' logger should work and there are special test functions which reference it by name. and this file doesn't use it anway.

mtp_api/apps/credit/views.py
```diff
@@ -258,15 +264,18 @@ class GetCredits(CreditViewMixin, mixins.ListModelMixin, viewsets.GenericViewSet
...
+        return q
 
     def get_serializer_class(self):
         if self.request.user.has_perm(
                 'transaction.view_bank_details_transaction'):
-            return SecurityCreditSerializer
+            return SecurityCreditCheckSerializer
         else:
-            return CreditSerializer
+            return CreditCheckSerializer
```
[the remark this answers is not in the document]
Staberinde: Yeah I wasn't sure if there were any hard or fast rules around who would get which permissions so I tried to implement it in the most flexible way that added the least amount of implicit logic

mtp_api/apps/credit/managers.py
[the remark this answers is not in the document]
Staberinde: Fair enough. I did want to bring up test assertions against log statements, and maybe talk about what value they bring to the tests. I feel like asserting against log statements and other ancillary developer-facing functionality leads you to testing functions rather than functionality; what do you think the chances of that test catching an unintentional removal of a log statement are vs catching an intentional removal/modification of that same log statement?

mtp_api/apps/credit/managers.py
```diff
@@ -65,6 +65,11 @@ def update_prisons(self):
                 (CREDIT_RESOLUTION.PENDING,)
             )
 
+    def create(self, *args, **kwargs):
+        credit_instance = super(CreditManager, self).create(*args, **kwargs)
+        credit_instance.attach_profiles()
```
ushkarev, line 70: tracing this method back to where it's called from for…
[transactions](https://github.com/ministryofjustice/money-to-prisoners-api/blob/0551de9982f0a9c13316603c18af2afc1b64c1e8/mtp_api/apps/transaction/serializers.py#L26) – this is fine, all the information a credit may have will already exist.
[payments](https://github.com/ministryofjustice/money-to-prisoners-api/blob/0551de9982f0a9c13316603c18af2afc1b64c1e8/mtp_api/apps/payment/serializers.py#L76) – i'm pretty sure this is too early. sender information is not known until the payment is captured so the sender profile will always be fixed as anonymous, won't it?

mtp_api/apps/credit/views.py
```diff
@@ -183,6 +189,8 @@ class CreditListFilter(django_filters.FilterSet):
 
     payment_reference = django_filters.CharFilter(field_name='payment__uuid', lookup_expr='startswith')
 
+    exclude_credit__in = NumberInFilter(field_name='id', lookup_expr='in', exclude=True)
```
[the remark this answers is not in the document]
Staberinde: Honestly not sure off the top of my head, I suspect `MultipleValueFilter` will probably produce a relatively similar test for membership as what we have here

mtp_api/apps/security/views.py
```diff
@@ -32,12 +32,12 @@
 from security.permissions import SecurityCheckPermissions, SecurityProfilePermissions
 from security.serializers import (
     AcceptCheckSerializer,
-    CheckSerializer,
+    CheckCreditSerializer,
     PrisonerProfileSerializer,
     RecipientProfileSerializer,
     RejectCheckSerializer,
     SavedSearchSerializer,
-    SenderProfileSerializer,
```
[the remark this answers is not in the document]
Staberinde: Apologies, I wouldn't normally change commas on imports unless I'd added an additional import and then removed it again. Happy to go with trailing commas as a rule if that's consistent across the codebase

mtp_api/apps/credit/managers.py
Staberinde: Ultimately it's going to be calling through to https://github.com/ministryofjustice/money-to-prisoners-api/blob/0551de9982f0a9c13316603c18af2afc1b64c1e8/mtp_api/apps/security/managers.py#L114 so it depends if the billing address on the payment will be populated at this point. If it's not caught here, then the `attach_profiles_for_legacy_credits` method on `update_security_profiles` will create the sender profile

mtp_api/apps/core/management/commands/load_test_data.py
```diff
@@ -172,3 +176,20 @@ def handle_prod(self, **options):
         ))
         verbosity = options.get('verbosity', 1)
         call_command('loaddata', 'initial_groups.json', verbosity=verbosity)
+
+    def create_super_admin(self):
```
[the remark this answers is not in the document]
Staberinde: Have made this change, will push it shortly

mtp_api/apps/core/management/commands/testserver.py
```diff
@@ -78,22 +77,8 @@ def load_test_data(self, verbosity=1, **kwargs):
...
-                first_name='Admin',
-                last_name='User',
-            )
-        for group in Group.objects.all():
-            admin_user.groups.add(group)
-        give_superusers_full_access()
-
-        self.stdout.write(self.style.SUCCESS('Model creation finished'))
+    def _create_super_admin(self):
+        return create_super_admin(self.stdout, self.style.SUCCESS)
```
ushkarev, line 81: you still need the `give_superusers_full_access` to be called after this to make sure the new super-admin has access to all client applications. otherwise the standalone test server won't have a usable admin user until something triggers the `load_test_data` management command.

ushkarev: the `create_super_admin` function itself totally does the correct thing tho

mtp_api/apps/security/serializers.py
```diff
@@ -218,7 +217,6 @@ def update(self, instance, validated_data):
 
 
 class CheckSerializer(serializers.ModelSerializer):
-    credit = SecurityCreditSerializer()
```
ushkarev, line 221: this line is still required as this is used unsubclassed in `security.views.CheckView`

ushkarev: wait no it doesn't anymore. what's the distinction between `CheckSerializer` and `CheckCreditSerializer`?

mtp_api/apps/security/management/commands/update_security_profiles.py
```diff
@@ -45,64 +46,104 @@ def handle_update(self, batch_size, recreate):
             PrisonerProfile.objects.update_current_prisons()
 
     def handle_credit_update(self, batch_size):
-        new_credits = Credit.objects.filter(sender_profile__isnull=True).order_by('pk')
-        new_credits_count = new_credits.count()
-        if not new_credits_count:
-            self.stdout.write(self.style.SUCCESS('No new credits'))
+        # TODO Remove below function once the logs show that it consistently
```
ushkarev, line 49: can't remove "legacy" function until bank transfers stop, right? maybe the TODO should say that explicitly

mtp_api/apps/security/serializers.py
Staberinde: `CheckCreditSerializer` serializes the attached credit, `CheckSerializer` does not. There are other ways to do this other than subclassing the serializer though, I can point you in the direction of a package that has this as one of it's core usecases. I didn't think it was worth adding just for this, but i'm happy to do so

ushkarev: oh i don't think we need any extra package just for this. i'm completely fine with serialiser subclassing!

at first i thought `CheckSerializer` was still being used in the check view, now it looks like it's never used directly except as the superclass of `CheckCreditSerializer`

mtp_api/apps/core/management/commands/load_test_data.py
```diff
@@ -32,8 +36,8 @@ class Command(BaseCommand):
     help = textwrap.dedent(__doc__).strip()
 
     def add_arguments(self, parser):
-        parser.add_argument('--protect-superusers', action='store_true',
-                            help='Prevents superusers from being deleted')
+        parser.add_argument('--no-protect-superusers', action='store_true',
```
ushkarev, line 39: there are references to `protect_superusers` in other places in the code, e.g.
https://github.com/ministryofjustice/money-to-prisoners-api/blob/master/mtp_api/apps/core/views.py#L203

would be safer to keep the variable name but swap the default by doing something like:
```
parser.add_argument('--no-protect-superusers', action='store_false', dest='protect_superusers' …
```

Staberinde: Sorry only just saw this comment. I've updated those other references now to be consistent with this negation for consistency. Let me know if you're happy with this, or if you still want me to go with what you describe above

mtp_api/apps/core/tests/utils.py
```diff
@@ -58,20 +60,24 @@ def make_applications():
...
 
     def make_application_and_roles(client_id, name, *roles):
-        new_app, _ = Application.objects.get_or_create(
-            client_id=client_id,
-            client_type='confidential',
-            authorization_grant_type='password',
-            client_secret=client_id,
-            name=name,
-            user=owner,
+        app = Application.objects.get(
```
ushkarev, line 63: won't this throw a `DoesNotExist`?

Staberinde: Yup, just ran into that issue in CI! Once again I was using a kludge of SQLAlchemy and Django ORM query syntax

PR #596, opened 2020-05-13
Run Check against created credit in load_test_data

mattbee:
To better test the FIU pending credits page, it would be useful to
have a set of checks to display on the pending credits page.

While this will create a bank of automatically accepted Checks, digging
through the test data to create checks that are placed in the delayed
capture state was proving difficult, so we can come back to that.

To make these Checks show in the credits to action page for FIU we can
update their state in Django admin for the time being.

### DIFF
--- a/mtp_api/apps/payment/tests/utils.py
+++ b/mtp_api/apps/payment/tests/utils.py
@@ -10,6 +10,7 @@ from faker import Faker
 from core.tests.utils import MockModelTimestamps
 from credit.constants import CREDIT_RESOLUTION
 from credit.models import Credit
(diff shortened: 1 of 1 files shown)


ushkarev:
i approved a bit too quickly :)

i think it’d be safer to do it in the admin form or view. this will have an effect on test data. i don’t think tests have been written with pre-existing checks (edited) 

all depends on whether the tests pass still!

ushkarev:
it could also slow down the whole test suite where checks are not necessary for the majority

mattbee:
Closing as this probably isn't going to work

PR #599, opened 2020-06-25
MTP-1452 Add/remove assignment from individual check page

Staberinde:
* Added `assigned_to` field on `security_check` 
* Added patch/put endpoint on `CheckView`
* Added tests for new behaviour
* Added `dictdiffer` to give more succinct diff on `assertCheckEqual` failure


mtp_api/apps/security/views.py
```diff
@@ -528,6 +528,7 @@ class Meta:
 class CheckView(
     mixins.ListModelMixin,
     mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
```
ushkarev, line 531: this enforces no restrictions on which user is set on `assigned_to` or what fields can be edited. this potentially leaves the api open to incorrect changes. e.g. if the status field is changed (by accident or whatever), then the proper business logic doesn't kick in.

do you think it might be worth having an "action" (like accept or reject below) that automatically assigns the user from the request? and (if need be) another to clear such assignment?

Staberinde: I mean, we could have seperate endpoints, but it seems more RESTful to have the `PATCH` method use the same url as the `GET`. What we could do though is stop `status` being mutated within the generic `update` method being used by the `CheckSerializer`? And for other fields that we don't want to mutate in any use case we could set them as `read_only` within the serializer?

Validation of the request against the user is another thing that could be done in the serializer (we've just pushed a commit that wasn't there when you reviewed that prevents assignment when someone else is assigned), just so we're encapsulating all the validation functionality all in one place. I'm tempted to leave out the validation of the request user, mainly because it would mean that if we ever made a page where the head of FIU could assign cases to her subordinates it would mean we could implement it without changes to the API repo. What do you think?

ushkarev: i agree this is more restful. yeah i think patch validation or read-only fields are a good idea.

mtp_api/apps/security/tests/test_views.py
```diff
@@ -1267,6 +1273,138 @@ def test_check_filtering_by_actioned_by(self):
...
+            data={
+                'assigned_to': assigned_to_user.id
+            },
+            format='json',
+            HTTP_AUTHORIZATION=auth,
+        )
+        self.assertEqual(response.status_code, http_status.HTTP_200_OK)
+
+        actual_check_data = response.json()
+        actual_check_data['assigned_to'] = assigned_to_user.id
```
ushkarev, line 1303: is this (and line below) not already set in the response? it seems a bit odd to issue a patch instruction and then modify the response as well to match… doesn't that mean this test will always pass because the response is manually tweaked?

mtp_api/apps/security/tests/test_views.py
```diff
@@ -1267,6 +1273,138 @@ def test_check_filtering_by_actioned_by(self):
...
+            data={
+                'assigned_to': new_assigned_to_user.id
+            },
+            format='json',
+            HTTP_AUTHORIZATION=auth,
+        )
+        self.assertEqual(response.status_code, http_status.HTTP_400_BAD_REQUEST)
+
+        # Assert check unchanged
+        actual_check_data['assigned_to'] = assigned_to_user.id
```
ushkarev, line 1346: same point as in the test above

mtp_api/apps/security/tests/test_views.py
```diff
@@ -1267,6 +1273,138 @@ def test_check_filtering_by_actioned_by(self):
...
+        self.assertEqual(response.status_code, http_status.HTTP_200_OK)
+
+        actual_check_data = response.json()
+        actual_check_data['assigned_to'] = new_assigned_to_user.id
+        actual_check_data['assigned_to_name'] = new_assigned_to_user.get_full_name()
+
+        check = Check.objects.get(pk=actual_check_data['id'])
+        self.assertCheckEqual(check, actual_check_data)
+
+
```
ushkarev, line 1407: can you pls add a test to double-check that another field cannot be patched? `status` in particular is the dangerous one.

mtp_api/apps/security/tests/test_views.py
Staberinde: Well this is embarrassing, that's supposed to be an assertion

PR #600, opened 2020-07-06
Improve documentation of requirements and local settings

ushkarev:


README.md
```diff
@@ -1,47 +1,56 @@
...
 
 ## Developing
 
 [![CircleCI](https://circleci.com/gh/ministryofjustice/money-to-prisoners-api.svg?style=svg)](https://circleci.com/gh/ministryofjustice/money-to-prisoners-api)
 
-Create a PostgreSQL database called `mtp_api`.
-
 It's recommended that you use a python virtual environment to isolate each application.
-Please call this `venv` and make sure it's in the root folder of this application so that `mtp_common.test_utils.code_style.CodeStyleTestCase` and the build tasks can find it.
+Please call this `venv` and make sure it's in the root folder of this application so that
```
Staberinde, line 15: I don't think it needs be called `venv`; I use the `virtualenvwrapper` helper which places the venvs in `~/.virtualenvs` and so far I haven't had any ill effects as I always start the venv beforehand

ushkarev: for the most part, it doesn't _have_ to be there. it's just recommended as a simple way to start. `mtp-common` does have some (not very strict) [references](https://github.com/ministryofjustice/money-to-prisoners-common/blob/2cacaf6d42a0ceccaa39cec28b9ece303c791939/mtp_common/test_utils/code_style.py#L29) to that location tho.

ushkarev: the idea is that if you know about virtual environments, you can do things your own way and ideally everything works nicely. but someone new who hasn't the experience can get started more easily

mattbee: Is it worth just rewording it so that it mentions why there might be an issue, and then, as you say, experienced people can proceed if they think they can handle any issues?

ushkarev: how about:


It's recommended that you use a python virtual environment to isolate each application.

The simplest way to do this is using:

```shell script
python3 -m venv venv    # creates a virtual environment for dependencies; only needed the first time
. venv/bin/activate     # activates the virtual environment; needed every time you use this app
```

Some build tasks expect the active virtual environment to be at `/venv/`, but should generally work regardless of
its location.

ushkarev: @Staberinde @mattbee ^^^

mattbee: That makes sense for me 👍

PR #601, opened 2020-07-07
MTP-1453 my list page

Staberinde:
https://dsdmoj.atlassian.net/browse/MTP-1453 
The only change to this PR is to add the `assigned_to` filter on `CheckListFilter`


ushkarev:
this pr is just about the last commit, right? i think it looks perfectly reasonable after a rebase

PR #602, opened 2020-07-14
Allow send-money app to read combined account balances from NOMIS

ushkarev:
[MTP-1465/MTP-1477](https://dsdmoj.atlassian.net/browse/MTP-1477)
- private cash, spends and savings account balances are combined into one total "combined_account_balance"
- balances are not available in private estate so 0 is returned; send-money will effectively ignore current balance
- errors from NOMIS are propagated to client

### DIFF
--- a/mtp_api/apps/prison/serializers.py
+++ b/mtp_api/apps/prison/serializers.py
@@ -88,6 +88,7 @@ class PrisonerValiditySerializer(serializers.ModelSerializer):
 
 
 class PrisonerAccountBalanceSerializer(serializers.Serializer):
+    NOMIS_ACCOUNTS = {'cash', 'spends', 'savings'}
     combined_account_balance = serializers.SerializerMethodField()
 
     def get_combined_account_balance(self, prisoner_location: PrisonerLocation):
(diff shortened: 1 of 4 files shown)


mtp_api/apps/prison/serializers.py
```diff
@@ -80,6 +87,34 @@ class Meta:
...
+
+        try:
+            nomis_account_balances = nomis.get_account_balances(
+                prisoner_location.prison.nomis_id,
+                prisoner_location.prisoner_number,
+            )
+            return (
+                nomis_account_balances['cash'] +
+                nomis_account_balances['spends'] +
+                nomis_account_balances['savings']
```
mattbee, line 106: Will Nomis values are always an integer, for instance for a new prisoner with no activity in savings for instance will definitely be 0?

ushkarev: the docs aren't explicit, but i've found nothing to suggest otherwise. at least we will immediately get an alert if an account is missing or has an unexpected type

Staberinde: Does the documentation specify anything about the types of these fields at all? If not, then can we add an explicit assertion on these types?

ushkarev: they say "All balance values are represented as pence values" with a sample response of `{  "cash": 13565,  "savings": 12344,  "spends": 5678}`. i'm happy to add some assertions

mtp_api/apps/prison/serializers.py
```diff
@@ -80,6 +87,37 @@ class Meta:
...
+                prisoner_location.prison.nomis_id,
+                prisoner_location.prisoner_number,
+            )
+            assert set(nomis_account_balances.keys()) == self.NOMIS_ACCOUNTS, 'response keys differ from expected'
+            assert all(
+                isinstance(nomis_account_balances[account], int) and nomis_account_balances[account] >= 0
+                for account in self.NOMIS_ACCOUNTS
+            ), 'not all response values are natural ints'
+        except AssertionError as e:
+            msg = f'NOMIS balances for {prisoner_location.prisoner_number} is malformed: {e}'
```
Staberinde, line 110: Might be worth putting the whole `nomis_account_balances` data structure in the message to make it easier to debug

mtp_api/apps/prison/tests/test_views.py
```diff
@@ -488,6 +491,99 @@ def test_valid_prisoner_not_found_with_incorrect_prison_filter(self):
         self.assertEmptyResponse(response)
 
 
+class PrisonerAccountBalanceTestCase(AuthTestCaseMixin, APITestCase):
```
Staberinde, line 494: Minor point, but given that the test setup for private and public prison test data seems mostly exclusive, and that no test method uses both, have you considered breaking this out into two test case classes?

mtp_api/apps/prison/serializers.py
ushkarev: yes i considered that, but we're supposed to keep our logs as clean of private information as possible. i figure that someone debugging would have the prisoner number so would actually make the call to NOMIS manually to see what was happening.

Staberinde:
What I have just realised though is that the fake data generators might need changing, as I set up the credit data generator used to create checks against to generate random values between £1 and £10k. Might be worth reducing the max limit to £1k (or maybe £100 as we only assign to 5 prisoners currently; a change I made to ensure there was enough useful decision history data to work with) to make the spread a bit more realistic

PR #603, opened 2020-07-20
Remove “Credit matched: ” prefix from security check description

ushkarev:
…so that the description can be appropriately labelled depending on where it is displayed.

[MTP-1488](https://dsdmoj.atlassian.net/browse/MTP-1488)

### DIFF
--- a/mtp_api/apps/security/migrations/0030_migrate_check_description.py
+++ b/mtp_api/apps/security/migrations/0030_migrate_check_description.py
@@ -1,0 +1,36 @@
+import django.contrib.postgres.fields
+from django.db import migrations, models
+
+
+SPLIT_CHECK_DESCRIPTION = '''
+UPDATE security_check
+SET description_migration = regexp_split_to_array(trim(description), '\\.\\s*');
+'''
+
+JOIN_CHECK_DESCRIPTION = '''
+UPDATE security_check
+SET description = array_to_string(description_migration, '. ');
+'''
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('security', '0029_remove_check_description_prefix'),
+    ]
+    operations = [
+        migrations.AddField(
+            model_name='check',
+            name='description_migration',
+            field=django.contrib.postgres.fields.ArrayField(base_field=models.CharField(max_length=100), blank=True, null=True, size=None),
+        ),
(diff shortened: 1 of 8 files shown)


Staberinde:
I'm happy enough with this change, but I feel like since we're going back and altering existing data, we should go further and change the data type of the field from a string to a Postgres `ARRAY` of strings, as it's a better representation of what the data actually is, and based off the discussion we were having at team review around how we might want to format this in the future I can see it as being potentially necessary. Data modelling changes are one of those code changes better done sooner rather than later, especially when they involve a data migration of a dataset that's only going to grow with time.

ushkarev:
interesting point @Staberinde maybe an array of descriptions is a good idea. we could split on the `.` because i don't think descriptions themselves have contained a full stop to date.

in the way it currently works, an array of descriptions matches the data model, as you say, in that there is one description line for each rule. but for the earliest checks, there was only one generic description for _multiple_ rules. still, splitting on `.` would work but there would be a different number of descriptions and matched rules for some checks. i don't think that's a real problem with that..? what do you think @Staberinde

Staberinde:
Yeah that all sounds reasonable to me, I think having a single description overlapping multiple rules for earlier checks isn't a problem, as we're only using this for rendering on the UI. And yes I think splitting on `.` is a sensible way to go about it (i'm agnostic as to whether you want to leave the `.` in place in each substring after the split)

mtp_api/apps/security/migrations/0030_migrate_check_description.py
```diff
@@ -0,0 +1,36 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('security', '0029_remove_check_description_prefix'),
+    ]
+    operations = [
+        migrations.AddField(
+            model_name='check',
+            name='description_migration',
+            field=django.contrib.postgres.fields.ArrayField(base_field=models.CharField(max_length=100), blank=True, null=True, size=None),
```
Staberinde, line 24: 100 characters sounds like quite a small limit for a rule, I can see this potentially causing issues in the future

mtp_api/apps/security/models.py
```diff
@@ -307,7 +307,11 @@ class Check(TimeStampedModel):
         choices=CHECK_STATUS,
         db_index=True,
     )
-    description = models.CharField(max_length=500, blank=True)
+    description = ArrayField(
```
Staberinde, line 310: `description` sounds a bit like it should still be a string

PR #608, opened 2020-08-20
MTP-1513 speed up circle ci

Staberinde:


.circleci/config.yml
```diff
@@ -1,150 +1,91 @@
 # generated by https://github.com/ministryofjustice/money-to-prisoners-deploy
```
ushkarev, line 1: it's worth removing this line as mtp-deploy doesn't yet generate these files

.circleci/custom_scripts/build_service_image.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/env sh
+. ${CIRCLE_WORKING_DIRECTORY}/.circleci/custom_scripts/pull_base_image.sh
```
ushkarev, line 2: i think it'd be more consistent with other scripts to put `common.sh` in this file. `pull_base_image.sh` could still be a separate file even though it's not use elsewhere

.circleci/config.yml
```diff
@@ -1,150 +1,91 @@
...
+        docker_layer_caching: true
+    - attach_workspace:
+        at: imagedump
+    - run:
+        name: Test docker image
+        command: . ${CIRCLE_WORKING_DIRECTORY}/.circleci/custom_scripts/test.sh
+    - run:
+        name: Extract test reports
+        when: always
+        command: |
```
ushkarev, line 40: maybe this should also be in a separate file so that _all_ shell scripts are files

ushkarev (commented):
pls can you also make the commit message more descriptive. it's not explaining the purpose of this change (which actually is not huge)

Staberinde:
I'd agree that some of the commit messages that have been squashed should be more descriptive. The purpose/business justifcation in full can be found via the Jira ticket id prepended to the commits.

.circleci/config.yml
Staberinde: I didn't think it was worth it in this case, as it's just 4 lines, none of which rely on dynamically templated environment variables. But I can move it out if you want

.circleci/custom_scripts/build_service_image.sh
Staberinde: Could do, but then it would be evaluated twice, once here and once in `pull_base_image.sh`

ushkarev: i would have either removed it from pull_base_image.sh or actually got rid of pull_base_image.sh since it's not reused elsewhere. that way, you have one script per step. it's no big deal really, just a tidiness suggestion

Staberinde: In my mind getting rid of it from `pull_base_image.sh` is a departure from consistency, and actually has a knock-on effect on functionality as then `pull_base_image.sh` can't be run by itself without erroring

ushkarev (approved):

PR #616, opened 2020-10-23
Allow for prisoner balances to be recorded internally as a fallback to checking NOMIS

ushkarev:
This will enable a work-around for capping balances in prisons which do not use NOMIS for prisoner monies (e.g. the entire private estate). Such prisons will provide a daily report to be uploaded into Django admin which will be used in place of a call to the NOMIS api.

[MTP-1641](https://dsdmoj.atlassian.net/browse/MTP-1641)


mtp_api/apps/prison/forms.py
```diff
@@ -1,6 +1,82 @@
...
+from prison.models import Prison, PrisonerBalance, validate_prisoner_number
+
 
 class LoadOffendersForm(forms.Form):
     modified_only = forms.BooleanField(label=_('Only load updated offender locations'), required=False)
+
+
+class PrisonerBalanceUploadForm(forms.Form):
+    csv_file = forms.FileField(label=_('Prisoner balances file'), allow_empty_file=True, widget=AdminFileWidget)
+    prison = ModelChoiceField(queryset=Prison.objects.all().order_by('-private_estate', 'nomis_id'))
```
mattbee, line 20: Now we have the `use_nomis_for_balances` field (added in a subsequent commit to this addition) - can we filter by that for the ChoiceField, (reducing room for error in selecting Birmingham essentially!)

ushkarev: yeah good idea @mattbee

mtp_api/apps/prison/forms.py
```diff
@@ -1,6 +1,82 @@
...
+        if csv_file:
+            try:
+                return self.parse_balances(csv_file)
+            except Exception as e:
+                self.add_error('csv_file', f'Could not parse file: {e}')
+        return None
+
+    @classmethod
+    def parse_balances(cls, csv_file):
+        csv_contents = csv_file.read().decode()
```
Staberinde, line 33: Are we using a custom `django.core.files.storage.Storage` subclass, or is `csv_file` just be a standard Python file object right? If it's the latter can't you just do `csv_reader = csv.DictReader(csv_file)`?

mtp_api/apps/prison/forms.py
```diff
@@ -1,6 +1,82 @@
...
+            if denomiator != 1:
+                raise ValueError(f'Cannot turn {amount} into pence')
+            balances.append({
+                'prisoner_number': prisoner_number,
+                'amount': int(amount_pence),
+            })
+        return balances
+
+    @transaction.atomic
+    def save(self):
```
Staberinde, line 57: It seems a bit strange to persist ORM objects on a form, maybe some kind of thin Serializer could be used?

mtp_api/apps/prison/forms.py
```diff
@@ -1,6 +1,82 @@
...
+        balances = [
+            PrisonerBalance(
+                prison=prison,
+                prisoner_number=balance['prisoner_number'],
+                amount=balance['amount'],
+            )
+            for balance in balances
+        ]
+        # delete all balances at this prison (because small balances will not be provided in spreadsheet)
+        delete_result = PrisonerBalance.objects.filter(prison=prison).delete()
```
Staberinde, line 69: Any reason why we couldn't combine these two delete queries into one?

mtp_api/apps/prison/migrations/0021_prison_use_nomis_for_balances.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 2.0.13 on 2020-10-23 11:15
+
+from django.db import migrations, models
+
+
+def flag_private_estate_as_not_using_nomis_for_balances(apps, schema_editor):
```
Staberinde, line 6: Are we setting Birmingham manually? If so which ticket is that a part of?

Staberinde (commented):
Looks good, just a couple of thoughts

mtp_api/apps/prison/forms.py
```diff
@@ -1,6 +1,82 @@
...
 
+from prison.models import Prison, PrisonerBalance, validate_prisoner_number
+
 
 class LoadOffendersForm(forms.Form):
     modified_only = forms.BooleanField(label=_('Only load updated offender locations'), required=False)
+
+
+class PrisonerBalanceUploadForm(forms.Form):
+    csv_file = forms.FileField(label=_('Prisoner balances file'), allow_empty_file=True, widget=AdminFileWidget)
```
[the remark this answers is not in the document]
ushkarev: we don't know how a truly empty file will look when exported from CMS. i think there's no harm in allowing it. we can use an empty file to clear all balances for a prison for example

mtp_api/apps/prison/forms.py
ushkarev: we tried that but the csv reader wants text and we couldn't get it to re-open the file in text mode (the special `file`-like object opens in binary mode)

mtp_api/apps/prison/forms.py
ushkarev: this is how the other file-uploading view does it. might as well use the same pattern, no? pretty sure that many forms in various apps actually do the db/api persistence inside them

mtp_api/apps/prison/forms.py
mattbee: I thought about that, but couldn't see how (might be lack of django knowledge). Can we filter and get results from both queries (prison OR prisoner_number__in) then might make sense.

mtp_api/apps/prison/migrations/0021_prison_use_nomis_for_balances.py
mattbee: No ticket, manual django admin task will probably be triggered by receiving a CSV from Birmingham if nothing else (changing the use_nomis_for_balance setting on a prison)

mtp_api/apps/prison/forms.py
ushkarev: no real reason other than it's probs a bit clearer as 2 calls. otherwise we'd have to do something like `.filter(Q(prison=prison) | Q(prisoner_number__in=set(balance.prisoner_number for balance in balances)))`… and i dunno if postgres would use the indices nicely

mtp_api/apps/prison/migrations/0021_prison_use_nomis_for_balances.py
ushkarev: birmingham is still unconfirmed. there's probably no specific ticket for it. there will likely be other manual setup – like the establishments which cannot provide these reports

mtp_api/apps/prison/forms.py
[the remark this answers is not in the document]
Staberinde: Maybe, my thinking was that I've had cases where my browser has decided to save a file with null bytes and I haven't realised till I've tried to open it. At the end of the day i'm not going to block over this though

mtp_api/apps/prison/tests/test_views.py
```diff
@@ -567,6 +568,38 @@ def test_retrieving_combined_balance_in_private_estate(self, mocked_nomis):
         self.assertDictEqual(response_data, {'combined_account_balance': 0})
         mocked_nomis.get_account_balances.assert_not_called()
 
+    @mock.patch('prison.serializers.nomis')
```
Staberinde, line 571: Might be nicer to use `RequestsMock` here to standardise how we mock external services across the coddebase(s), but i'm not going to block over this

Staberinde (approved):
Nothing worth blocking over

PR #617, opened 2020-10-26
MTP-1512 Tolerate failure of NOMIS account balance requests

Staberinde:


ushkarev (approved):
all looks good in essence! there are a couple of missing line breaks that flake8 pointed out

we ought to keep an eye out for 404s in case they come up unexpectedly instead of a 400 for a person not at a prison we expected. (i'm not suggesting something in the code)

ushkarev (approved):

PR #618, opened 2020-11-09
Parse most commonly supplied field name for prisoner balance CSV import

ushkarev:
The initial sample we had used "prisonernumber" but "prisonnumber" is actually being supplied by most prisons.


kjdchapman (approved):
> most prisons

are some prisons still giving us `prisonernumber`?

ushkarev:
@kjdchapman not one of them uses `prisonernumber`, some use neither however (cos they're editing it themselves)

PR #619, opened 2020-11-13
Fix exporting of very old debit card credits for Analytical Platform

ushkarev:
…from when billing address was not recorded


ushkarev:
there's no urgency to this at all. it should probably go in after the django 2.2 stuff

Staberinde (approved):

PR #622, opened 2020-12-18
MTP-1654 Check value not key for rejection reasons

Staberinde:


ushkarev (approved):

ushkarev (approved):

ushkarev:
with the fix in noms-ops being deployed, i think this PR is no longer necessary. i'd prefer the api to reject payloads that don't logically make sense with the action. if you still do want to merge it, fine i guess :)

ushkarev (approved):

PR #623, opened 2021-01-08
Fix searching for prisoner balances in django admin

ushkarev:
Name is not available, this should have searched by number


xoen (approved):
👍

PR #625, opened 2021-01-15
Simpler python dependency tree

ushkarev:
Simplify python dependency tree by installing the correct version of mtp-common on _first run_ (i.e. when none is installed). This subsumes the dependencies for a production/deployed image into the base requirements which also simplifies testing.

Depends on [mtp-common#401](https://github.com/ministryofjustice/money-to-prisoners-common/pull/401)

### DIFF
--- a/run.py
+++ b/run.py
@@ -3,8 +3,8 @@ if __name__ == '__main__':
     import os
     import sys
 
-    if sys.version_info[0:2] < (3, 4):
-        raise SystemExit('python 3.4+ is required')
+    if sys.version_info[0:2] < (3, 6):
+        raise SystemExit('Python 3.6+ is required')
 
     root_path = os.path.abspath(os.path.dirname(__file__))
 
--- a/run.py
+++ b/run.py
@@ -6,24 +6,28 @@ if __name__ == '__main__':
     if sys.version_info[0:2] < (3, 4):
         raise SystemExit('python 3.4+ is required')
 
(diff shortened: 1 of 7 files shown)


xoen (approved):
LGTM 👍

run.py
```diff
@@ -3,27 +3,31 @@
...
             raise SystemExit('setuptools and pip are required')
+        try:
+            pip = pkg_resources.load_entry_point('pip', 'console_scripts', 'pip')
+        except pkg_resources.ResolutionError:
+            raise SystemExit('setuptools and pip are required')
 
-        print('Pre-installing MTP-common')
-        pip_main(['--quiet', 'install', '--upgrade', 'money-to-prisoners-common'])
+        print('Pre-installing MTP-common and base requirements')
+        pip(['install', '--requirement', f'{root_path}/requirements/base.txt'])
```
Staberinde, line 28: Do we know that ` -r <file>` is equivalent to `--upgrade --upgrade-strategy=only-if-needed -r <file>` for both the upgrade and downgrade case?

ushkarev: i believe upgrades and downgrades are treated the same by `pip`, simply as a version mismatch

ushkarev: but this pip invocation doesn't try to upgrade installed dependencies that satisfy the requirements files. i.e. it's
```
pip install --upgrade-strategy=only-if-needed -r <file>
```
not
```
pip install --upgrade --upgrade-strategy=only-if-needed -r <file>
```

PR #626, opened 2021-01-18
MTP-1738 API auto-accept functionality

Staberinde:
Ticket link: https://dsdmoj.atlassian.net/browse/MTP-1738
Used by: https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/510
To discuss:
* What to do on auto-accept POST/PATCH failure. Do we revert the state of the check, otherwise they won't be able to make another attempt at creating an auto-accept rule


mtp_api/apps/security/managers.py
```diff
@@ -420,3 +434,16 @@ def _get_matching_rules(self, credit):
             if rule.applies_to(credit) and rule.triggered(credit):
                 matched_rule_codes.append(rule_code)
         return matched_rule_codes
+
+
+class CheckAutoAcceptRuleManager(models.Manager):
+
+    def get_active_auto_accept_for_credit(self, credit: Credit):
+        auto_accept_rule = self.filter(
+            debit_card_sender_details=credit.sender_profile.debit_card_details.first(),
```
xoen, line 443: I don't know the models very well so apologies if it's a stupid question.

Is it possible that any of the related models in this chain `credit.sender_profile.debit_card_details` is not present? and you think we should guard/check against that? 

But maybe all of them will be present in the case of a credit paid via debit card so maybe this is fine, but worth asking ☺️

mtp_api/apps/security/models.py
```diff
@@ -376,6 +383,44 @@ def __str__(self):
...
+        return self.get_latest_state().active
+
+    class Meta:
+        ordering = ('-created',)
+        unique_together = (
+            ('debit_card_sender_details', 'prisoner_profile',),
+        )
+
+
+class CheckAutoAcceptRuleState(TimeStampedModel):
```
xoen, line 409: Do you reckon is it also worth adding a default ordering by `created DESC` for the states? It seems like the natural orders, e.g. latest has priority over older states.

mtp_api/apps/security/migrations/0034_auto_20210114_1234.py
```diff
@@ -0,0 +1,25 @@
...
+        migrations.AddField(
+            model_name='check',
+            name='auto_accept_rule',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='checks', to='security.CheckAutoAcceptRule'),
+        ),
+        migrations.AlterField(
+            model_name='check',
+            name='rejection_reasons',
+            field=django.contrib.postgres.fields.jsonb.JSONField(default=dict),
+        ),
```
xoen, line 24: Didn't the `Check` model have a `rejection_reasons` field before?

Maybe it's a leftover from another migration/branch?

mtp_api/apps/security/serializers.py
```diff
@@ -349,3 +352,76 @@ def reject(self, by):
...
+
+
+class CheckAutoAcceptRuleSerializer(serializers.ModelSerializer):
+    states = CheckAutoAcceptRuleStateSerializer(many=True, required=True)
+
+    def validate(self, attrs):
+        if len(attrs['states']) != 1:
+            raise ValidationError(
+                f'When creating or updating an auto-accept rule, states must be of length 1, not {len(attrs.states)}'
+            )
```
xoen, line 381: Interesting, so you changed the serializer and it's now getting a `states` array rather than a `reason`.

So if I understand this correctly you got rid of the abstraction and now don't try to hide the rule/states implementation details but this means the client needs to pass `states` explicitly.

So:
1. on `create`, `states` would have the initial state. This also means that potentially the client can create a non-active state
2. on `update`, `states` only has the new state which is added

Is that correct?

mtp_api/apps/security/serializers.py
```diff
@@ -349,3 +352,76 @@ def reject(self, by):
...
+        return super().validate(attrs)
+
+    def create(self, validated_data):
+        auto_accept_rule = CheckAutoAcceptRule.objects.create(
+            debit_card_sender_details=validated_data['debit_card_sender_details'],
+            prisoner_profile=validated_data['prisoner_profile'],
+        )
+        CheckAutoAcceptRuleStateSerializer().create(
+            validated_data={
+                'active': True,
```
xoen, line 391: Oh interesting, the create request may have `"active": false` but we still create the first state with `active=True`. 

Do you reckon is worth adding a validation if the `active` field is passed as `False` on create?

mtp_api/apps/security/tests/test_checks.py
```diff
@@ -557,3 +557,141 @@ def test_pending_check_created_for_monitored_user(self):
...
+            reverse(
+                'security-check-auto-accept-list'
+            ),
+            data={
+                'prisoner_profile': prisoner_profiles[0].id,
+                'debit_card_sender_details': sender_profiles[0].debit_card_details.first().id,
+                'states': [
+                    {
+                        'reason': 'This person has amazing hair',
+                        'added_by_id': self.users['security_fiu_users'][0].id
```
xoen, line 585: Does the client still needs to pass `states.added_by_id` if in the serializer [we set it from the context's `request.user`](https://github.com/ministryofjustice/money-to-prisoners-api/pull/626/files#diff-3cf42ec26dec1e69b6d44a59bb6095460400d603136ea39c825fd17666fa2a86R393)? e.g. is `states.added_by_id` a mandatory field?

Maybe it's because of how serializers work.

mtp_api/apps/security/tests/test_views.py
```diff
@@ -1820,3 +1826,274 @@ def test_cannot_reject_an_accepted_check(self):
...
+            reverse(
+                'security-check-auto-accept-list'
+            ),
+            data={
+                'prisoner_profile': self.prisoner_profile.id,
+                'debit_card_sender_details': self.debit_card_sender_details.id,
+                'states': [
+                    {
+                        'reason': 'they have amazing hair',
+                        'added_by': self.added_by_user.id
```
xoen, line 1873: Client may pass this `added_by` field but it's effectively ignored right?

mtp_api/apps/security/tests/test_views.py
```diff
@@ -1820,3 +1826,274 @@ def test_cannot_reject_an_accepted_check(self):
...
+            'states': [
+                {
+                    'active': True,
+                    'reason': 'they have amazing hair',
+                    'added_by': {
+                        'first_name': self.added_by_user.first_name,
+                        'last_name': self.added_by_user.last_name,
+                        'username': self.added_by_user.username,
+                    }
+                },
```
xoen, line 2001: I'm wondering if the "when" the state changed/was created will be needed later on.

...but this is probably a problem for another ticket ☺️

mtp_api/apps/security/views.py
```diff
@@ -586,3 +588,21 @@ def reject(self, request, pk):
...
+    viewsets.GenericViewSet,
+):
+    queryset = CheckAutoAcceptRule.objects.all()
+    filter_backends = (DjangoFilterBackend, filters.OrderingFilter)
+    serializer_class = CheckAutoAcceptRuleSerializer
+    ordering_fields = ('created',)
+    ordering = ('created',)
+    permission_classes = (
+        IsAuthenticated,
+        SecurityCheckPermissions,
```
xoen, line 606: So we're reusing the `SecurityCheckPermissions` permission class here.

I'm not that familiar with this, would this use the `CheckAutoAcceptRule`'s permissions because `SecurityCheckPermissions` inherits from `ActionsBasedPermissions` (through `SecurityProfilePermissions`)?

xoen (dismissed):
Nice work @Staberinde 

I left few maybe annoying comments/questions but it makes sense to me with my limited knowledge ☺️ 

It may still be worth for someone else with more knowledge of everything to have a look as well maybe.

mtp_api/apps/security/serializers.py
```diff
@@ -220,9 +224,92 @@ def update(self, instance, validated_data):
         return super().update(instance, validated_data)
 
 
+class CheckAutoAcceptRuleStateSerializer(serializers.ModelSerializer):
+    active = serializers.BooleanField(required=False)
```
ushkarev, line 228: why is this not required? is it just to allow this call to omit the `active` param?
https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/510/files#diff-8ea59d9e31bda3ec134036535c9ff9709e0d74eb87794d9b602a87fb2314a638R345
might make the code and validation errors a bit clearer if on create it had to be True, but present always.

mtp_api/apps/security/serializers.py
```diff
@@ -220,9 +224,92 @@ def update(self, instance, validated_data):
...
+        ]
+        read_only_fields = (
+            'id',
+            'auto_accept_rule'
+        )
+
+    def validate(self, attrs):
+        if self.instance and attrs.get('active') is None:
+            raise ValidationError({
+                'active': ValidationError(_('On update, states[0].active must be specified'))
```
ushkarev, line 248: related to the above point about why `active` isn't always required. also with this version, you can post True two successive times.

mtp_api/apps/security/tests/test_checks.py
ushkarev: the field on the serialiser is `added_by` and i'm not sure a related model field would be magical enough to also try a `added_by_id` param in the data. i suspect this line has no effect.

mtp_api/apps/security/tests/test_views/test_check_views.py
```diff
@@ -0,0 +1,1231 @@
...
+
+        actual_check_data = response.json()
+        self.assertCheckEqual(check, actual_check_data)
+
+    def test_get_check_auto_accept_rule_state_attached(self):
+        """
+        Test that the get object endpoint returns CheckAutoAcceptRuleState when appropriate
+        """
+        # Setup
+        response = self.client.post(
```
ushkarev, line 598: to me, it's a little self-referential to use a view to create test data in order to test other views. in other places we either use the model or mommy creation methods directly. but really i don't mind since this data-creating view itself is tested elsewhere in isolation from other views.

mtp_api/apps/security/tests/test_views/test_check_views.py
```diff
@@ -0,0 +1,1231 @@
...
+                        'added_by': self.added_by_user.id
+                    }
+                ]
+            },
+            format='json',
+            HTTP_AUTHORIZATION=self.auth,
+        )
+        self.assertEqual(response.status_code, 201)
+        actual_response = response.json()
+        self.assertIn('id', list(actual_response.keys()))
```
ushkarev, line 1012: just fyi, you can simply do `self.assertIn('id', actual_response)` because dicts have a `__contains__` method that checks for a key's presence

mtp_api/apps/security/serializers.py
```diff
@@ -220,9 +224,92 @@ def update(self, instance, validated_data):
...
+        read_only_fields = [
+            'id',
+            'created',
+        ]
+
+
 class CheckSerializer(serializers.ModelSerializer):
     actioned_by_name = serializers.SerializerMethodField('get_actioned_by_name_from_user')
     assigned_to_name = serializers.SerializerMethodField('get_assigned_to_name_from_user')
+    auto_accept_rule_state = CheckAutoAcceptRuleStateSerializer(read_only=True, required=False)
```
ushkarev, line 312: is it worth _always_ having this present? i know we need this for the auto-accept review page and we need the _latest_ state probably always. see, i'm just thinking of the credits to action and history pages which will be loading additional models for every row and then not displaying them. then again, we are not expecting a huge number of auto-accepts anyway. just thinking "out loud"

mtp_api/apps/security/tests/test_views/test_check_views.py
```diff
@@ -0,0 +1,1231 @@
...
+                'states': [
+                    {
+                        'active': True,
+                        'reason': 'they grew it back again',
+                        'added_by': self.added_by_user.id
+                    }
+                ]
+            },
+            format='json',
+            HTTP_AUTHORIZATION=self.auth,
```
ushkarev, line 1207: is it a pain to ask for one of these to be for `self.users['security_fiu_users'][1]`? just so we can prove that multiple users' details propagate

ushkarev (commented):
just some relatively minor comments!

mtp_api/apps/security/tests/test_views.py
ushkarev: it'll certainly be needed for the auto-accept history view. as you say that is a logical iteration for the future

mtp_api/apps/security/managers.py
ushkarev: it doesn't really make sense for a credit to have a check made without first having a sender profile and debit card details. i think this is good.
see `payment.models.create_security_check_if_needed_and_attach_profiles`

mtp_api/apps/security/serializers.py
ushkarev: see point just below that's related. i'm not saying this needs to change or anything.

how i understand it works now is:
- new auto-accept does require an `active` flag because the serialiser always uses True
- subsequent calls require an `active` flag
- but they do not validate whether it's the opposite of the latest flag

we could validate subsequent calls to ensure the flag cycles True/False, but i feel that's overkill. i'd probably have just made it simpler by making `active` always but overridden it to True on creation.

mtp_api/apps/security/serializers.py
Staberinde: It's an internal API which is only used by a single client, I think it's fine for the API to create the initial `AutoAcceptRule` and `AutoAcceptRuleState` in the proper state without using the value for active given in the request. There's no reason to create an `AutoAcceptRule` in the `active=False` state, and if for some reason the POST endpoint was called for an existing check, then a database integrity error would be returned because of the unique constraint (`debit_card_details_id` and `prisoner_profile_id`) defined on the model.

mtp_api/apps/security/tests/test_checks.py
Staberinde: So this is an interesting one, I believe this behavior actually changed in Django 2.2, which we've recently upgraded to. Now when `ModelSerializer` serializes a model with a foreign key and no explicit nested serializer defined, it strips off the '_id' suffix of the field. I think it's backwards compatible on deserialization, but not on serialization (i.e. it will accept the previous behavior of the `_id` suffix, but won't add it when serializing it and any other objects to the response.

I'll rename the `added_by_id` key,  to `added_by` for consistency though, as that's what the actual call that's being made attaches as a parameter.

ushkarev: oh ok. still, you shouldn't need to explicitly pass it in as noms-ops doesn't and the serialiser automatically gets it from the request

mtp_api/apps/security/tests/test_views.py
Staberinde: It will, the `CheckAutoAcceptRuleState.created` field is serialized as part of the `CheckAutoAcceptRuleStateSerializer` and asserted for existence in the API tests.

mtp_api/apps/security/models.py
Staberinde: Thinking about it, i'm not sure that `created DESC` makes any sense outside of the context of `get_latest_state`. Adding `created` ordering on `CheckAutoAcceptRuleState` and changing `CheckAutoAcceptRule` from `created DESC` to `created` sounds good though.

mtp_api/apps/security/serializers.py
```diff
@@ -220,9 +224,92 @@ def update(self, instance, validated_data):
...
+                _(f'When creating or updating an auto-accept rule, states must be of length 1, not {len(attrs.states)}')
+            )
+        return super().validate(attrs)
+
+    def create(self, validated_data):
+        auto_accept_rule = CheckAutoAcceptRule.objects.create(
+            debit_card_sender_details=validated_data['debit_card_sender_details'],
+            prisoner_profile=validated_data['prisoner_profile'],
+        )
+        CheckAutoAcceptRuleStateSerializer().create(
```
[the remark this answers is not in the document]
Staberinde: It's just a useful pattern for creating related objects in django to enforce separation of concerns. It has the added advantage that any validation (which is just going to be for the presence of the `reason` field in this case) is done by the nested serializer, and any validation error is properly nested in the response.

mtp_api/apps/security/serializers.py
```diff
@@ -220,9 +224,92 @@ def update(self, instance, validated_data):
...
+
+class CheckAutoAcceptRuleSerializer(serializers.ModelSerializer):
+    states = CheckAutoAcceptRuleStateSerializer(many=True, required=True)
+
+    def validate(self, attrs):
+        if len(attrs['states']) != 1:
+            raise ValidationError(
+                _(f'When creating or updating an auto-accept rule, states must be of length 1, not {len(attrs.states)}')
+            )
+        return super().validate(attrs)
```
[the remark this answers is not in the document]
Staberinde: It actually already works in the way that you describe: http://www.cdrf.co/3.9/rest_framework.serializers/ModelSerializer.html#run_validation

mtp_api/apps/security/serializers.py
Staberinde: `active` is not required on `create` but is required everywhere else. The view will use these rules for its validation, even those in the nested serializer, so the data structure received to the API needs to conform to these rules (to prove this. comment out this line and run the tests). I agree that it would be better to be explicit about the conditions when it is needed, I'll codify the above into a  `verify_active` method.

Agreed about not validating the flipping of state. I'll leave the actual mutation of the values as it is if you're happy with that

ushkarev: i'm perfectly happy either way!

mtp_api/apps/security/serializers.py
Staberinde: Yeah, so I had originally intended on keeping the abstraction, but Django Rest Framework Serializers don't really facilitate expecting a differing data structure on `create`, `update` and serialization, at least not in a nice way, so I moved away from that. I'm happy to make the endpoint structure more explicit with respect to what's happening underneath (i.e. removing the `UpdateModelMixin` from `CheckAutoAcceptRuleView`, and creating a new `AutoAcceptRuleState` endpoint that's create only and has the `CheckAutoAcceptRule.id` as a positional URL parameter, and then removing `CheckAutoAcceptSerializer.update` in favour of the default `CheckAutoAcceptState.create` implementation from `ModelSerializer`. That's probably a more RESTful way of approaching the problem.

Staberinde: Ah, I remember why I didn't do this originally; there's not really a nice way to ensure that `CheckAutoAcceptRuleSerializer.create` tolerates the addition of the `reason` parameter for the writable case; for serializing it's simple enough to do this using `serializers.SerializerMethodFields` and a `get_*` method for the read case, but no equivalent exists for the write case. I really don't want to break the creation up into API calls into multiple API calls as that's just asking for data integrity problems.

ushkarev: yeah i agree. it's easiest to just keep it all together; the logic is not complex

mtp_api/apps/security/serializers.py
Staberinde: So as I understand it the history page will eventually need this information, as it's used to determine whether a check has been auto-accepted rather than just accepted (as they are to be displayed differently). Credits to action is a fair point, I could break this out into another serializer so that it doesn't get loaded for that case, but we already have a confusing hierarchy of check serializers and I don't want to make it worse. Ideally we'd want a way of selectively choosing which relations to serialize (a company I used to work for made a library for just this use case https://django-rest-framework-serializer-extensions.readthedocs.io/en/latest/) but i'd see that as a seperate refactoring ticket(s) as there's quite a few places we could implement it

mtp_api/apps/security/tests/test_views/test_check_views.py
Staberinde: Not at all, it's a fair point

mtp_api/apps/security/tests/test_views/test_check_views.py
Staberinde: I see what you're saying, my thinking is that it's not worth adding other test setup factory that we're only going to use in this class, and if we're going to set up the data somehow, then it's it's nice and explicit to use the actual API

mtp_api/apps/security/tests/test_views/test_check_views.py
Staberinde: Fair point, do you want me to make that change here?

ushkarev: up to you really! i mean this works fine

ushkarev (approved):

ushkarev (dismissed):
looks good. we'll have to see what the others think with regards to the TODO comment. i'm not too set on either route. perhaps i slightly lean towards adding multiple "active" states as there's more evidence preserved for users to see what happened.

mtp_api/apps/security/tests/utils.py
```diff
@@ -82,34 +97,74 @@ def _get_credit_values(credit_filters, sender_profile_id, prisoner_profile_id, p
     )
 
 
+def _get_prisoner_profile(
+    j: int, filter_set: dict, number_of_prisoners_to_use: int, fake_prisoner_names: dict, fiu
+) -> Tuple[Optional[PrisonerProfile], str]:
```
ushkarev, line 102: shouldn't the typing hint be `Tuple[Optional[…], Optional[str]]` for these methods?

ushkarev: or even more strictly `Union[Tuple[…, str], Tuple[None, None]]`

Staberinde: I've made the first change, I don't know if we gain much with the second but i'm happy to implement if you're keen

ushkarev (approved):

xoen:
FWIW I had another skim and didn't see any blockers 👍

PR #628, opened 2021-01-29
admin: Improve disbursements search

xoen:
By allowing user to search disbursements also by NOMIS transaction ID
or by Invoice Number.

In addition to these fields being searched, the Invoice Number is now
shown in the table as well as this is useful.

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1774


ushkarev (approved):

xoen:
Thanks for the review @ushkarev 👍

PR #629, opened 2021-02-01
admin: Credit details, make owner and private_estate_batch R/O

xoen:
These fields were editable which meant when seeing a Credit details
they were pulling extra data to populate the selects.

This page is used in a read-only fashion so these fields are now read-only
to speed-up the loading time of this page.

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1800


ushkarev (approved):

xoen:
Thanks for the review @ushkarev ☺️

PR #630, opened 2021-02-03
Remove code related to Single Offender ID

xoen:
The project was dropped and the code never really used so we can clean things a bit, reduce noise and probably avoid potential future confusion.

**Overview of changes**:
- from a DB standpoint, the `single_offender_id` field/column is removed from the [`Credit`](https://github.com/ministryofjustice/money-to-prisoners-api/pull/630/commits/91aca6737e74b8b17d73bf8e4dff650a75d8886f), [`PrisonerLocation`](https://github.com/ministryofjustice/money-to-prisoners-api/pull/630/commits/bedb08a6a3fb8db70212b7f5be0d09526164bbd4) and [`PrisonerProfile`](https://github.com/ministryofjustice/money-to-prisoners-api/pull/630/commits/366e02d5dd9c7e9bcc944ed6f9fd5db07c18ebf2) models.
- the other notable change is the [removal of the `load_prisoner_locations` command](https://github.com/ministryofjustice/money-to-prisoners-api/commit/6f7539998acc7d6c2c40f8502b17cd255a1c99fd)
- this command invoked in the `LoadOffendersView` ('Load offenders' admin button) but this functionality is not really used
- furthermore the settings for the Single Offender ID service endpoint/credentials (`OFFENDER_API_*`) can now go as well
- [some test utils](https://github.com/ministryofjustice/money-to-prisoners-api/pull/630/commits/c6bcec5745aac83f4364dfb91703114f8e443604) were also setting these `single_offender_id` which is not gone, as a result a couple of these utility functions used to generate data are slightly simpler

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1672

PR #631, opened 2021-02-04
Removing SOID migrations (so that these columns are not deleted)

xoen:
The `single_offender_id` column was removed from 3 tables/models.

This may cause problems in production so to we're thinking a good
approach to this would be to delete these migrations (so that the
columns are not dropped in production) and once this is released to
production a separate PR with these migrations added back can be
released when no code is using these fields.

Related to PR: https://github.com/ministryofjustice/money-to-prisoners-api/pull/630

Slack conversation: https://mojdt.slack.com/archives/C015WEVRQSX/p1612439976010100?thread_ts=1612438739.002800&cid=C015WEVRQSX


ushkarev (approved):

ushkarev:
you didn't make this on top of master, but can still force-merge

xoen:
@ushkarev oh because I didn't rebase since the merge, sorted 👍

Staberinde (approved):

PR #633, opened 2021-02-05
CI: Run Django checks and check if code misses migrations

xoen:
Running some Django checks as part of our CI build pipeline:
- `./manage.py checks` and
- `./manage.py makemigrations --check` which would return a non-zero status when new migrations were generated. e.g. potentially we forgot to run `makemigrations` and we didn't include them

This additional `run_checks` job will run alongside the `test` job as it should be relatively fast and it's useful to know as soon as possible if we missed something in a PR for example.

**NOTE**: Also note that this would not block push of the docker image or deploy to test.
The assumption is that sometimes it may make sense to temporarily leave some DB schema changes/migrations out of a PR.

If we later on decide this checks should be blocking we can always change the CI workflow to be more draconian.

Screenshot 2021-02-05 at 16 02 17

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1796

### DIFF
--- a/.circleci/custom_scripts/start_db_container.sh
+++ b/.circleci/custom_scripts/start_db_container.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env sh
+
+docker run \
+  --name postgres \
+  --detach \
+  -e POSTGRES_USER=postgres \
+  -e POSTGRES_PASSWORD=postgres \
+  -e POSTGRES_DB=mtp_api \
+  circleci/postgres:10.10-alpine
+
+sleep 10
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,17 @@ jobs:
         path: /tmp/reports
     - store_test_results:
         path: /tmp/reports
+  run_checks:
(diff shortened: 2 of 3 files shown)


.circleci/custom_scripts/start_db_container.sh
```diff
@@ -8,4 +8,4 @@ docker run \
   -e POSTGRES_DB=mtp_api \
   circleci/postgres:10.10-alpine
 
-sleep 10
+ps_isready --timeout 10 --host postgres
```
ushkarev, line 11: this won't work as it's outside the container with the `circleci/postgres:10.10-alpine` image

xoen: as I misspelled the binary name I'm pretty sure this wouldn't be in that image :P 

but more seriously, I see what you mean, I give it a quick try but I'll drop the commit 👍

xoen: (attempt/commit removed)

ushkarev (approved):

PR #635, opened 2021-02-09
Fix migration

ushkarev:
to match change made in commit: aabb048cb409a9622496c15baee1862212ce5173

This has no effect on the database fields in postgres so a fix-up to an old migration is valid; it's not necessary to make a new migration and deploy.

So now CI ~should be~ is happy that migration files match models.


ushkarev:
Screenshot 2021-02-09 at 14 47 41
CI seems happy now

xoen (approved):
👍

PR #636, opened 2021-02-22
MTP-1428: Fix missing month in Staff Logins report

xoen:
This mysterious bug caused the disappearing of March in the Staff Login
report table header.

### Problem

On Sunday 29 March 2020 the clocks went forward to get more daylight
(this is called Daylight Saving Time, or DST for short).
This started the period of the year called British Summer Time (BST) ending
the Greenwich Mean Time (GMT) period:

- 1 March 2020 GMT (UTC +00:00)
- 1 April 2020 BST (UTC +01:00)

Because of the way we were calculating the beginning of the current month
and the previous 3 ones, by using the current day (timezone aware)
and by replacing the month component, that datetime object retained the
same timezone info, including the offset from UTC time.

Unfortunately, 1st of March and 1st of February are not in BST but
are in GMT which means when converted to localtime using the wrong offset
they'll be converted to 29 February (23:00) and 31st January (23:00)
respectively.

This is why the user, in May, would see April, February and January as
3 previous months (instead of the correct April, March, February).

### Solution

The solution is to use the desired year/month to construct "naive" datetimes
(datetimes without an associated timezone) and then convert these into
"aware" datetimes (datetimes with an associated timezone) using
`django.utils.timezone.make_aware()`.

`make_aware()` will cleverly use the correct timezone and offset.

### Refactorings

- Moved "get current month progress" logic into the
 `get_current_month_progress()` method.
  This was in `get_months()` to get previous 3 months, but these are
  separate concerns so it makes more sense to separate them.
- use variables for the "seconds from beginning of month" and "seconds in
  the month" concepts to hopefully improve readability.

### Links
UK clock change: https://www.gov.uk/when-do-the-clocks-change

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1428

### DIFF
--- a/mtp_api/apps/mtp_auth/tests/test_login_counts.py
+++ b/mtp_api/apps/mtp_auth/tests/test_login_counts.py
@@ -68,3 +68,83 @@ class LoginCountTestCase(TestCase):
         self.assertEqual(login_counts[(None, this_month)], 2)
         self.assertEqual(login_counts[(prison.nomis_id, last_month)], 3)
         self.assertEqual(login_counts[(None, last_month)], 0)
+
+    def test_get_months(self):
+        scenarios = [
+            {
+                'name': 'no edge cases',
+                'now': make_aware(datetime.datetime(2020, 11, 15, 12)),
+                'expected_months': [
+                    make_aware(datetime.datetime(2020, 11, 1)),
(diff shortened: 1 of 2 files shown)


ushkarev (approved):

xoen:
Thanks for the review @ushkarev!

PR #643, opened 2021-03-12
Fixed duplicate parameter error in swagger without impacting production functionality

Staberinde:
Whilst developing against MTP-1805, I realized the `CheckAutoAcceptRuleView` endpoint had broken swagger because there were two `ordering` parameters; one provided by `filters.OrderingFilter` and one provided by `CheckAutoAcceptRuleFilter`. Since the former can't filter on related fields, and was being overridden by the latter I removed the former.

I also added a simple test to spot this issue if it occurs again. I put it in `core/tests/` for lack of a better place, but i'm open to moving it

### DIFF
--- a/mtp_api/apps/core/tests/test_swagger.py
+++ b/mtp_api/apps/core/tests/test_swagger.py
@@ -1,0 +1,12 @@
+
+from rest_framework.test import APITestCase
+
+
+class TestSwagger(APITestCase):
+
+    def test_get(self):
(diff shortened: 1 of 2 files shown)


ushkarev (approved):
i like the idea of test-loading the swagger docs.

makes me think it's worth doing that in general, even for boring non-dynamic pages. i might put that into the tech backlog (low priority methinks)

PR #645, opened 2021-03-16
Stop posting stats to GOV.UK Performance Platform

ushkarev:
…because GDS have killed it off. The functionality for posting stats remains, but is no longer scheduled to run periodically.

### DIFF
--- a/api.ini
+++ b/api.ini
@@ -41,5 +41,4 @@ spooler-chdir = %d
 spooler-import = mtp_%n/tasks.py
 cron = 0 3 -1 -1 -1 %d/venv/bin/python %d/manage.py clean_up
 cron = -1 -1 -1 -1 -1 %d/venv/bin/python %d/manage.py run_scheduled_commands
-cron = 30 13 -1 -1 -1 %d/venv/bin/python %d/manage.py check_tokens
 # attach-daemon = %d/venv/bin/python %d/manage.py load_data_listener
--- a/mtp_api/apps/performance/management/commands/update_performance_platform.py
+++ b/mtp_api/apps/performance/management/commands/update_performance_platform.py
@@ -1,3 +1,7 @@
+"""
+NB: GOV.UK have retired the performance platform
+"""
+
 from datetime import datetime
 
 from django.core.management import BaseCommand
--- a/mtp_api/apps/performance/migrations/0004_retire_performance_platform.py
+++ b/mtp_api/apps/performance/migrations/0004_retire_performance_platform.py
@@ -1,0 +1,33 @@
+from django.db import migrations
+
+
+def remove_performance_platform_update(apps, schema_editor):
+    cls = apps.get_model('core', 'ScheduledCommand')
+    cls.objects.filter(name='update_performance_platform').delete()
+
+
+def restore_performance_platform_update(apps, schema_editor):
+    cls = apps.get_model('core', 'ScheduledCommand')
+    cls.objects.create(
+        name='update_performance_platform',
+        arg_string='--resources completion-rate',
+        cron_entry='0 5 * * Mon',
+    )
+    cls.objects.create(
+        name='update_performance_platform',
+        arg_string='--resources transactions-by-channel-type',
+        cron_entry='50 22 * * *',
+    )
+
+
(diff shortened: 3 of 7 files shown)


mtp_api/apps/performance/migrations/0004_retire_performance_platform.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def remove_performance_platform_update(apps, schema_editor):
+    cls = apps.get_model('core', 'ScheduledCommand')
+    cls.objects.filter(name='update_performance_platform').delete()
+
+
+def restore_performance_platform_update(apps, schema_editor):
+    cls = apps.get_model('core', 'ScheduledCommand')
+    cls.objects.create(
```
ushkarev, line 11: these 2 are here because i wanted there to be a record _somewhere_ of the current scheduled tasks

api.ini
```diff
@@ -41,5 +41,4 @@ spooler-chdir = %d
 spooler-import = mtp_%n/tasks.py
 cron = 0 3 -1 -1 -1 %d/venv/bin/python %d/manage.py clean_up
 cron = -1 -1 -1 -1 -1 %d/venv/bin/python %d/manage.py run_scheduled_commands
-cron = 30 13 -1 -1 -1 %d/venv/bin/python %d/manage.py check_tokens
```
xoen, line 44: I haven't checked out this branch and looked at the whole code on my machine (and still looking at this PR) but what is this `check_tokens` task and is it related to the performance platform?

mtp_api/apps/performance/management/commands/update_performance_platform.py
```diff
@@ -1,3 +1,7 @@
+"""
+NB: GOV.UK have retired the performance platform
```
xoen, line 2: worth deleting this whole file? (and `mtp_api/apps/performance/updaters.py`) or are they still used somehow?

xoen (dismissed):
Asked some questions but LGTM 👍

api.ini
ushkarev: it's not related to the performance platform stuff, but i mention in 2cf2230ff53a2af0aa42bd4694516ac2ba9332a5 that the management command had already been removed in 98f920fef289cff2781c7ed81384f26524a30b09

mtp_api/apps/performance/management/commands/update_performance_platform.py
ushkarev: as i say, i'm only suggesting removing the scheduling of these updates (which raise errors) rather than deleting the updaters altogether at this point.

api.ini
xoen: Oh sorry, I didn't look at the individual commits and I missed that explanation - makes sense 👌

mtp_api/apps/performance/view_dashboard.py
```diff
@@ -15,11 +15,17 @@
...
-    ).json()
+    try:
+        response = requests.get(
+            'https://www.performance.service.gov.uk/data/send-prisoner-money/customer-satisfaction?'
+            'flatten=true&duration=1&period=year&collect=rating_1%3Asum&collect=rating_2%3Asum&collect=rating_3%3Asum&'
+            'collect=rating_4%3Asum&collect=rating_5%3Asum&collect=total%3Asum&format=json'
+        )
+        response.raise_for_status()
+    except requests.RequestException:
+        return 'Not available'
```
xoen, line 26: Does the client code use this returned value or it's just ignored/displayed directly?

If the client code doesn't do anything "special" with this (e.g. use as a key for a dictionary or what not) this should be fine.

xoen (approved):
LGTM 👍

mtp_api/apps/performance/view_dashboard.py
xoen: AFAICT this is set in a view context and displayed directly in the template so it should be fine 👌

ushkarev: i did check, it just gets displayed in the template as-is. i used this pattern because, a few lines down, this function can return a string as an error message if no data was available

PR #646, opened 2021-03-17
Fix asset building and include latest dependencies from mtp-common

ushkarev:
The most important part of this PR is upgrading to mtp-common 11.3: 11.x before this one assumed that apps use a single-tree structure for building assets (scripts and stylesheets) whereas `api` actually keeps them in separate folders and does not make bundles.

In future, we want to have the api make asset bundles in the same fashion as other apps so it is not treated a special case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@ node_modules/
 npm-debug.log
 
 # app-specific
+.eslintrc.json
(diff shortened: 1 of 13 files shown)


xoen (approved):
LGTM 👍

PR #647, opened 2021-03-22
MTP-1824 Allow FIU to manage all security users

Staberinde:
Jira: https://dsdmoj.atlassian.net/browse/MTP-1824

TODO:
* [x] Should we constrain against adding `FIU` group without `UserAdmin` group in Django Admin?
    * My inclination is yes, so that we constrain against bad data entry by someone outside the prisoner money team
 * [x] Add upgrade script to add `UserAdmin` group to all users with `FIU` group
   * [x] Are we also removing `UserAdmin` from any `Security` users without `FIU` group at the same time?


mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -811,7 +823,12 @@ def assertUserCreated(self, requester, user_data, client_id, groups,  # noqa: N8
             set(new_user.groups.all()),
             set(groups)
         )
-        self.assertSamePrisons(requester, new_user, msg='User Admin able to retrieve users without matching prisons')
+        if assert_prisons_inherited:
+            self.assertSamePrisons(
+                requester, new_user, msg='User Admin able to retrieve users without matching prisons'
+            )
+        else:
+            self.assertNoPrisons(requester, new_user)
```
ushkarev, line 831: this is checking that the `requester` user has no prisons. should it not be the new user that's verified? something like `self.assertNoPrisons(new_user, msg='New user should have no prisons attached')`

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -1474,6 +1536,52 @@ def test_cannot_update_prison_set_for_cashbook_user(self):
...
+        self.assertUserUpdated(requesting_user, user.username, user_data)
+        updated_user = User.objects.get(username=user.username)
+        self.assertSequenceEqual(
+            updated_user.groups.order_by('name').values_list('name', flat=True),
+            ['FIU', 'Security', 'UserAdmin']
+        )
+        updated_prisons = PrisonUserMapping.objects.get_prison_set_for_user(
+            updated_user
+        ).all()
+        self.assertSequenceEqual(list(current_prisons), list(updated_prisons))
```
ushkarev, line 1560: do you think it's better to compare these as sets in case ordering is unstable for some reason?

mtp_api/apps/mtp_auth/tests/test_views.py
```diff
@@ -1474,6 +1536,52 @@ def test_cannot_update_prison_set_for_cashbook_user(self):
...
+        self.assertUserUpdated(requesting_user, user.username, user_data)
+        updated_user = User.objects.get(username=user.username)
+        self.assertSequenceEqual(
+            updated_user.groups.values_list('name', flat=True),
+            ['Security']
+        )
+        updated_prisons = PrisonUserMapping.objects.get_prison_set_for_user(
+            updated_user
+        ).all()
+        self.assertSequenceEqual(list(current_prisons), list(updated_prisons))
```
ushkarev, line 1583: same point as above about stable ordering

mtp_api/apps/mtp_auth/tests/test_views.py
Staberinde: Can do, IIRC this should be a list of a single element though. My reasoning for using this is that it was what was used in the associated tests

ushkarev: from what i could tell, the other list/sequence comparisons specified a sort within the test so hopefully should be good. i might well have missed some tho!

mtp_api/apps/mtp_auth/serializers.py
```diff
@@ -181,11 +178,18 @@ def create(self, validated_data):
         new_user.set_password(password)
         new_user.save()
 
-        if make_user_admin:
+        role.assign_to_user(new_user)
+        # In custom flow, enforce that `UserAdmin` is added to any new `Security` user who has the `can_manage` box
+        # ticked in the form.
+        # Note that the django admin user creation flow does not use this endpoint so this logic won't apply there
+        if make_user_admin or new_user.groups.filter(name='FIU').exists():
```
xoen, line 185: I'm reading this and I'm wondering if this `if` statement could be split into two `if` statements to make things more clear.

AFAICT (and I may well be wrong!), if `make_user_admin=True` then we want to add this user to the `UserAdmin` group - this is basically the logic which was there before.

On top of that, if the user is part of `FIU` group then we also want to add them to the `UserAdmin` group.

And finally, if someone is in the `Security` group and can manage users (is in the `UserAdmin` group) that means they should be added to the `FIU` group, that seems to be implicit by the fact this `if` is nested within this `if` - but again, maybe I'm misunderstanding this.

What I'm wondering and potentially suggest is something like:

```Python
# FIU users can manage users
if new_user.groups.filter(name='FIU').exists():
  make_user_admin = True

if make_user_admin:
  new_user.groups.add(Group.objects.get(name='UserAdmin'))

# Security users that can manage users should also be in the FIU group
if make_user_admin and new_user.groups.filter(name='Security').exists():
  new_user.groups.add(Group.objects.get(name='UserAdmin'))
```

I suppose it's the same logic but maybe it's a bit easier to follow 🤷

ushkarev: i'm not sure that's clearer. bearing in mind that the driving input to decide whether somebody is an admin _or_ FIU is the `user_admin=True` from the POST request. to me, it's more logical to have that drive the decision to add both FIU and UserAdmin groups

ushkarev: what i mean is, to me it's clearer if only the POST request sets `make_user_admin` and this variable doesn't then change. but it's subjective, right? it's only about readability rather than overall effect.

xoen: (Thinking aloud here) Actually I also notice that this logic to add/remove a 'Security' user to/from the 'FIU' group is repeated in several places and I do wonder if it would beneficial to move into a function to sync those two groups.

Maybe something like this:

```Python
def sync_security_fiu_groups(user):
    """
    Add/remove a 'Security' user to/from the 'FIU' group.

    If a user is in the 'Security' group and can manage users they should
    also be in the 'FIU' group.

    Conversely, if a 'Security' user can't manage users they should be
    removed from the 'FIU' group.
    """

    can_manage_user = updated_user.groups.filter(name='UserAdmin').exists()
    is_security = updated_user.groups.filter(name='Security').exists()

    if is_security:
        if can_manage_user:
            user.groups.add(Group.objects.filter(name='FIU').first())
        else:
            user.groups.remove(Group.objects.filter(name='FIU').first())
```

You could then simplify some of the code below with something like:

```python
if make_user_admin:
  updated_user.groups.add(user_admin_group)
else:
  updated_user.groups.remove(user_admin_group)

sync_security_fiu_groups(updated_user)
```
(or something like that, all of this is pretty much pseudo-code and I may miss a lot of things here)

but maybe it's overkill and it would hide this logic and make these create/update methods more difficult to read 🤷

ushkarev: yeah that works for me, but i have no strong opinions either way :)

xoen: @ushkarev 

> i'm not sure this code snippet is clearer. bearing in mind that the driving input to decide whether somebody is an admin or FIU is the user_admin=True from the POST request. to me, it's more logical to have that drive the decision to add both FIU and UserAdmin groups

Yes, I see what you mean about the input variable being repurposed 😬 - it was a bit of an idea and wrote that way to avoid the `or` in the first `if ` and/or to avoid the repetition in of the `new_user.groups.add()`, you may well write that like:

```python
# FIU users can also manage users
if make_user_admin or new_user.groups.filter(name='FIU').exists():
  new_user.groups.add(Group.objects.get(name='UserAdmin'))

# Security users that can manage users should also be in the FIU group
if make_user_admin and new_user.groups.filter(name='Security').exists():
  new_user.groups.add(Group.objects.get(name='UserAdmin'))
```

I suppose the thing I struggle a little to read was the nested if to add a `Security` user to the `FIU` group but maybe it's just me have a preference with flat `if` statements without and/or (whenever it's possible of course) ☺️

ushkarev: we definitely cannot have 

```python
if make_user_admin and new_user.groups.filter(name='Security').exists():
  new_user.groups.add(Group.objects.get(name='UserAdmin'))
```

because then no users outside of the security tool can be made an admin

xoen: Yep 👍, sorry that was a copy&paste mistake which I noticed when I re-read my comment, it was meant to be add to the 'FIU' group clearly ☺️

Staberinde: So i'm happy to break it out into functions if it makes sense to do so. 

Couple of points worth making though:

Regarding your first comment, I think having nested if statements in this case is clearer than having shallow if statements predicated on two conditions. Also I'm not sure if  redefining a value we're receiving as input to the form adds to the clarity of the code; as Igor says I think it's clearer to use the notion of the `user_admin` provided by the form here. Also the `FIU` group wouldn't be ever present on the user at this point; as FIU doesn't have an existing role, it fits in the existing functionality much better if we think of them as admins of the `Security` role, so that the only change we need to make to the generic code flow as it stands is to make sure that any user created with `role=security` and `make_admin=True` has the `FIU` group added to them, rather than the other way round (see my comment https://dsdmoj.atlassian.net/browse/MTP-1818?focusedCommentId=143381)

> ```python
> def sync_security_fiu_groups(user):
>     """
>     Add/remove a 'Security' user to/from the 'FIU' group.
> 
>     If a user is in the 'Security' group and can manage users they should
>     also be in the 'FIU' group.
> 
>     Conversely, if a 'Security' user can't manage users they should be
>     removed from the 'FIU' group.
>     """
> 
>     can_manage_user = updated_user.groups.filter(name='UserAdmin').exists()
>     is_security = updated_user.groups.filter(name='Security').exists()
> 
>     if is_security:
>         if can_manage_user:
>             user.groups.add(Group.objects.filter(name='FIU').first())
>         else:
>             user.groups.remove(Group.objects.filter(name='FIU').first())
> ```

I think this last line will raise an exception when updating a normal `Security` user, as the `FIU` group won't be part of that users group set

xoen (dismissed):
FWIW this LGTM 👍 

Left some comments/ideas but if you and Igor think what's there is clear enough I don't mind ☺️

mtp_api/apps/mtp_auth/serializers.py
ushkarev: i agree with you @Staberinde 

on a side note, `user.groups.remove(Group.objects.filter(name='FIU').first())` can be turned into `user.groups.filter(name='FIU').remove()` (or is it .delete()?) so that it doesn't raise exceptions

mtp_api/apps/mtp_auth/tests/test_views.py
Staberinde: Good spot, I hate the way python interprets additional positional arguments as keyword arguements regardless of keyword

ushkarev (approved):

mtp_api/apps/mtp_auth/migrations/0017_Remove_UserAdmin_Group_from_Security_if_not_FIU.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.2.17 on 2021-03-31 15:44
+
+from django.db import migrations
+
+def remove_UserAdmin_group_from_security_if_not_FIU(apps, schema_editor):
+    user_model = apps.get_model('auth', 'user')
+    group_model = apps.get_model('auth', 'group')
+    security_users = user_model.objects.filter(groups__name='Security').exclude(groups__name='FIU').all()
```
ushkarev, line 8: should we exclude super-admins? i.e. us :)

mtp_api/apps/mtp_auth/serializers.py
```diff
@@ -152,40 +149,63 @@ def validate(self, attrs):
...
+                attrs['role'] = managed_roles[role]
             except KeyError:
                 raise serializers.ValidationError({'role': 'Invalid role: %s' % role})
 
         prisons = self.initial_data.get('prisons')
         if prisons is not None:
-            prison_objects = Prison.objects.filter(
-                nomis_id__in=[prison['nomis_id'] for prison in prisons]
-            )
+            if all([isinstance(prison, Prison) for prison in prisons]):
```
ushkarev, line 158: so `self.initial_data['prisons']` is sometimes dicts and sometime `Prison`s? how does the latter situation come about?

mtp_api/apps/mtp_auth/migrations/0017_Remove_UserAdmin_Group_from_Security_if_not_FIU.py
Staberinde: Probably, although being a part of the `UserAdmin` group wouldn't make any difference to our ability to create/edit users in Django admin as we are granted that through the inbuilt superadmin toggle. 

But it's maybe something to think about as a wider point; I added a question about that here https://github.com/ministryofjustice/money-to-prisoners-api/pull/647/files to remind myself to talk about it, but then forgot :)

mtp_api/apps/mtp_auth/serializers.py
Staberinde: `AccountRequestView.partial_update` loads the prison as part of `get_queryset` https://github.com/ministryofjustice/money-to-prisoners-api/blob/8cf0c9f45584fde3898672584d1ed4b317a206f0/mtp_api/apps/mtp_auth/views.py#L397

xoen:
Hi @Staberinde FYI I've just merged [the other PR](https://github.com/ministryofjustice/money-to-prisoners-api/pull/649) related to the account requests work. You may want to rebase this one on top of `master` ☺️

mtp_api/apps/mtp_auth/views.py
```diff
@@ -465,14 +465,19 @@ def partial_update(self, request, *args, **kwargs):
             serializer_kwargs = {}
             user_existed = False
 
+        if instance.prison:
+            prisons = [instance.prison]
+        else:
+            prisons = None
```
xoen, line 471: 👍

mtp_api/apps/mtp_auth/serializers.py
```diff
@@ -225,25 +247,35 @@ def update(self, user, validated_data):
             updated_user.groups.clear()
 
             if make_user_admin:
-                updated_user.groups.add(user_admin_group)
+                self._make_user_admin(updated_user)
```
xoen, line 250: I just spotted that this is causing a little bug when updating a 'Security' user to be a `UserAdmin`/FIU: It wouldn't add them to FIU group as expected.

This is a consequence of using `_make_user_admin()` _just after_ their groups are cleared above (`updated_user.groups.clear()`) as this helper method checks is user is part of `Security` group but at that point they'll belong to no groups.

```python
    def _make_user_admin(new_user):
        new_user.groups.add(Group.objects.get(name='UserAdmin'))
        # We add the FIU group to enforce current requirement that Security UserAdmin's are synonymous with FIU
        if new_user.groups.filter(name='Security').exists():
            new_user.groups.add(Group.objects.get(name='FIU'))
```

Move this call after the role's groups are added back below:

```python
            for group in role.groups:
                updated_user.groups.add(group)
```

should solve this issue...

...but I'll probably fix in a separate PR after this PR is merged.

xoen (approved):
Had another look and LGTM 👍

mtp_api/apps/mtp_auth/serializers.py
xoen: FYI: Fixed in this other PR: https://github.com/ministryofjustice/money-to-prisoners-api/pull/652

PR #650, opened 2021-04-13
MTP-1863 Remove remaining identifiers from exceptions and attach to sentry metadata

Staberinde:
Jira: https://dsdmoj.atlassian.net/browse/MTP-1863

* Moved dynamic fields out of logger.error/warning and exceptions to allow for sentry to deduplicate

Note that i've left any exceptions propagate to the UI/API and do not touch Sentry alone (`ValidationError`, `InvalidStateForUpdateException` etc)

Depends on [common#419](https://github.com/ministryofjustice/money-to-prisoners-common/pull/419)

### DIFF
--- a/mtp_api/apps/core/templatetags/tick_charts.py
+++ b/mtp_api/apps/core/templatetags/tick_charts.py
@@ -33,7 +33,7 @@ def get_zero(value):
         return datetime.timedelta()
     if isinstance(value, (int, float)):
         return 0.0
-    raise ValueError('Cannot determine zero value for a %s' % type(value))
+    raise ValueError('Cannot determine zero value', {'type': type(value)})
 
 
 @register.inclusion_tag('core/tick-chart.html')
--- a/mtp_api/apps/credit/management/commands/create_prisoner_credit_notices.py
+++ b/mtp_api/apps/credit/management/commands/create_prisoner_credit_notices.py
@@ -37,11 +37,11 @@ class Command(BaseCommand):
         try:
             prison = Prison.objects.get(pk=prison)
         except Prison.DoesNotExist:
-            raise CommandError('Prison %s does not exist' % prison)
+            raise CommandError('Prison does not exist', {'prison_nomis_id', prison})
         if date:
             date = parse_date(date)
             if not date:
-                raise CommandError('Date %s cannot be parsed, use YYYY-MM-DD format' % date)
+                raise CommandError('Date cannot be parsed, use YYYY-MM-DD format', {'date_string': date})
         else:
             date = now().date() - datetime.timedelta(days=1)
         date_range = (make_aware(datetime.datetime.combine(date, datetime.time.min)),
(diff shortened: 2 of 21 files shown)


mtp_api/apps/core/templatetags/tick_charts.py
```diff
@@ -33,7 +33,7 @@ def get_zero(value):
         return datetime.timedelta()
     if isinstance(value, (int, float)):
         return 0.0
-    raise ValueError('Cannot determine zero value for a %s' % type(value))
+    raise ValueError('Cannot determine zero value', {'type': type(value)})
```
xoen, line 36: I wonder if add the value to the Sentry context would also be useful to debug/investigate.
I tend to prefer having a bit more information, as it helps with debugging problems but obviously this is not wrong or anything ☺️

mtp_api/apps/credit/management/commands/create_prisoner_credit_notices.py
```diff
@@ -37,11 +37,11 @@ def handle(self, path, prison, date=None, **options):
         try:
             prison = Prison.objects.get(pk=prison)
         except Prison.DoesNotExist:
-            raise CommandError('Prison %s does not exist' % prison)
+            raise CommandError('Prison does not exist', {'prison_nomis_id', prison})
```
xoen, line 40: Does raising a `CommandError` exception behaves like using `logger.error()` in the sense that this other parameter ends up being in the Sentry context?

mtp_api/apps/transaction/views.py
```diff
@@ -74,8 +74,10 @@ def patch_processed(self, request, *args, **kwargs):
         try:
             return self.partial_update(request, *args, **kwargs)
         except InvalidCreditStateException as e:
-            logger.warning('Some transactions failed to update: [%s]' %
-                           ', '.join(map(str, e.conflict_ids)))
+            logger.warning(
+                'Some transactions failed to update',
+                {'transaction_id_list': e.conflict_ids}
```
xoen, line 79: I'm curious, this list is going to be serialised into a string like `[1, 2, 3]` on Sentry right? Or maybe it just stays a list there?

xoen (dismissed):
I've added some comments/questions but this looks good to me 👍

mtp_api/apps/transaction/views.py
Staberinde: I doubt Sentry will transform the data type, but I haven't validated that

mtp_api/apps/credit/management/commands/create_prisoner_credit_notices.py
ushkarev: it'd appear in a similar way in sentry, but the title of the error is just the exception class name (`CommandError` in this case) with the description being a stringified version of all arguments (`('Prison does not exist', {'prison_nomis_id', ????})` in this case)

xoen (approved):

PR #651, opened 2021-04-19
'send_account_request_emails' works with requests without prison

xoen:
'security' account requests are not prison-specific.
These 'security' account requests are emailed to all the 'security' `UserAdmins`
(FIU), regardless of which prisons are associated to these `UserAdmins`.

The test for this command have been improved to separate the test for the
`find_admins()` method from the test which checks correct admins are emailed.

**NOTE**: The `mtp_auth/new_account_requests.(html|txt)` templates don't seem
to use the `prison_name` in the context so I've removed it.


ushkarev (approved):

xoen:
Thanks @ushkarev - FYI, test for this command was failing because of a change in the test FIU users, so I've updated this test now: https://github.com/ministryofjustice/money-to-prisoners-api/pull/651/commits/797edfa532c99ea86690204662ee19e25a3b8c0c

ushkarev (approved):

PR #654, opened 2021-05-04
Handle SatisfactionDashboard failing requests

xoen:
Some of the surveys and/or questions have gone now and this dashboard raises
an exception while trying to get this data:

https://sentry.io/organizations/ministryofjustice/issues/2139335407/?environment=prod&project=2098210

This is caused by the fact the HTTP response is probably a non-200
but we try to get the data out of the response body anyway.

I've added `requests`' `raise_for_status()` and handling the requests
exception by logging them and by not adding the question to the dashboard
results

### DIFF
--- a/mtp_api/apps/core/dashboards.py
+++ b/mtp_api/apps/core/dashboards.py
@@ -1,5 +1,6 @@
 import json
 import itertools
+import logging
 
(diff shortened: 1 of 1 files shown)


ushkarev (approved):
👍

PR #658, opened 2021-05-10
mtp-common 12.0.0: Only support Python 3.8+

xoen:
As part of the migration to a newer version of Ubuntu (`20.04`) which
comes with Python 3.8.

Related to ticket: https://dsdmoj.atlassian.net/browse/MTP-1841

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,6 +1,6 @@
 # Dependencies needed for all environments
 
-money-to-prisoners-common~=11.4.0
+money-to-prisoners-common~=12.0.0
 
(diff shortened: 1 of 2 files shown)


ushkarev (changes requested):
you should also change the requirement in `dev.txt`

xoen:
Oh I see sorry, I thought `dev.txt` inherited from `base.txt`, it does but it install a `[testing]` version of `mtp-common` so the version is specified there as well.

Thanks, I'll change 👌

ushkarev (approved):

PR #660, opened 2021-05-19
Added various DB indexes to improve performances 

xoen:
Some of the DB tables's columns are part of searched fields in various filters but had no DB indexes.
This change aim to improve the performances in general but specifically aim to [improve the speed of search of credits in `noms-ops`](https://dsdmoj.atlassian.net/browse/MTP-1273).

## Indexes added

Column(s) | Why?
------------ | -------------
[`Credit.prisoner_name`](https://github.com/ministryofjustice/money-to-prisoners-api/commit/5b0616dbd3ff6af218218c5f4c2e0b5cf2cb1e4f) | [Filtered in `CreditListFilter`](https://github.com/ministryofjustice/money-to-prisoners-api/blob/20453b0a4013924313a485e64b88e49fefaf42c8/mtp_api/apps/credit/views.py#L69) via [`CreditTextSearchFilter`](https://github.com/ministryofjustice/money-to-prisoners-api/blob/20453b0a4013924313a485e64b88e49fefaf42c8/mtp_api/apps/credit/views.py#L176)
[`Payment.cardholder_name`, `Payment.email`, `Payment.card_expiry_date`, `Payment.card_number_first_digits` and `Payment.card_number_last_digits`](https://github.com/ministryofjustice/money-to-prisoners-api/commit/4cbdefae84ae519f2d4f482358cccc0d2864e9af)| [`Payment.cardholder_name` and `Payment.email` are fields searched in the Credit 'simple search'](https://github.com/ministryofjustice/money-to-prisoners-api/blob/20453b0a4013924313a485e64b88e49fefaf42c8/mtp_api/apps/credit/views.py#L170-L171). The others are [searched in the more advanced search](https://github.com/ministryofjustice/money-to-prisoners-api/blob/20453b0a4013924313a485e64b88e49fefaf42c8/mtp_api/apps/credit/views.py#L195-L197)
[`Payment.recipient_name`](https://github.com/ministryofjustice/money-to-prisoners-api/commit/27018181011c7855d56a3f857b2f7b428555c9ae) | [Searched in `PaymentAdmin`](https://github.com/ministryofjustice/money-to-prisoners-api/blob/20453b0a4013924313a485e64b88e49fefaf42c8/mtp_api/apps/payment/admin.py#L70)
[`Transaction.sender_name`, `Transaction.sender_account_number`, `Transaction.sender_roll_number` and `Transaction.sender_sort_code`]()https://github.com/ministryofjustice/money-to-prisoners-api/pull/660/commits/660d570bbad97909eeff88d8ea65bf4a89dfb903 | `Transaction.sender_name` is [searched as part of the Credits 'simple search'](https://github.com/ministryofjustice/money-to-prisoners-api/blob/20453b0a4013924313a485e64b88e49fefaf42c8/mtp_api/apps/credit/views.py#L169). The others are [part of the more advanced credits search](https://github.com/ministryofjustice/money-to-prisoners-api/blob/20453b0a4013924313a485e64b88e49fefaf42c8/mtp_api/apps/credit/views.py#L182-L188) (admittedly these are less relevant now that bank transfers' transactions are not used anymore, but I've added these indexes anyway)
[`BillingAddress.postcode`](https://github.com/ministryofjustice/money-to-prisoners-api/pull/660/commits/4efdae2ed26a15c66ed923da57e825e8f40121f7) | field [searched in credits search](https://github.com/ministryofjustice/money-to-prisoners-api/blob/20453b0a4013924313a485e64b88e49fefaf42c8/mtp_api/apps/credit/views.py#L199)

Hopefully these indexes will also benefit other queries/part of the applications (probably Django admin filtering too)

## On adding indexes without locking the DB table

Adding a DB index to a table is an operation which locks the table. It's tricky to predict how long add an index will take.

PostgreSQL has the [ability to create an index concurrently](https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) without locking the table while the index is created. This has the benefit of not causing downtime while the index is created.

Django [introduced `AddIndexConcurrently` in version 3.x](https://docs.djangoproject.com/en/3.2/releases/3.0/), but we're currently using Django 2.x.

This means create an index concurrently needs to be done running SQL (see this very well written [article on the subject](https://realpython.com/create-django-index-without-downtime/)):
- because `CREATE INDEX CONCURRENTLY` cannot be run within a DB transaction, these migration have `atomic = False`
- as we run SQL via `RunSQL`, in order to avoid confusing Django causing the creation of more migrations we're using `SeparateDatabaseAndState` to tell Django what's the desired state after the migration run
- as we run SQL via `RunSQL`, `reverse_sql='DROP INDEX "..."` is passed so that these migrations can be rolled back

**NOTE**: In the case of the indexes added to the `transaction_transaction` table I've not used this technique and used the default Django migration. The Transaction table is not updated as frequently these days as we don't accept bank transfers anymore and it's relatively small compared to Checks/Credit/Payment with few hundreds rows - so I've kept this simple.

## Tickets
- generic: https://dsdmoj.atlassian.net/browse/MTP-1523
- credit search: https://dsdmoj.atlassian.net/browse/MTP-1273


xoen:
**NOTE**: After further investigation we found that these indexes are not used by the query to search credits because it's mostly looking for values containing the keyword:

https://www.postgresql.org/docs/10/indexes-types.html

Longer explanation with further links in [this comment on the ticket](https://dsdmoj.atlassian.net/browse/MTP-1523?focusedCommentId=154303).

**Closing this PR** (we could always revisit/or pick part of this later on)

CC: @ushkarev

PR #661, opened 2021-05-21
Turn on Sentry.io performance monitoring

ushkarev:
…at 20% sample rate in production (this is a very high traffic app) and 100% in other environments.


xoen (approved):
👍

PR #665, opened 2021-06-08
Improve data exporting tool

ushkarev:
This will allow the export of data to be more configurable and provide more information to the Analytical Platform.
Foundational step to implement export to Linkspace #666 / [MTP-1873](https://dsdmoj.atlassian.net/browse/MTP-1873)

Analytical Platform will now also get data about:
• credits that were rejected by FIU (and why)
• who is monitored currently by FIU
• relationships that are auto-accepted without manual FIU checks

### DIFF
--- a/mtp_api/apps/credit/dump.py
+++ b/mtp_api/apps/credit/dump.py
@@ -1,0 +1,108 @@
+from django.conf import settings
+
+from core.dump import Serialiser
+from credit.constants import CREDIT_RESOLUTION, CREDIT_STATUS
+from credit.models import Credit, LOG_ACTIONS as CREDIT_LOG_ACTIONS
+from payment.models import PAYMENT_STATUS, BillingAddress
+from security.models import CHECK_STATUS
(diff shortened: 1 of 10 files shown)


mtp_api/apps/credit/dump.py
```diff
@@ -0,0 +1,159 @@
...
+                )
+            else:
+                security_check_rules = None
+                security_check_rejection_reasons = None
+        else:
+            security_check_description = None
+            security_check_status = None
+            security_check_actioned_by = None
+            security_check_rules = None
+            security_check_rejection_reasons = None
```
xoen, line 94: I wonder if logic to serialise the `security_check` could be swept in a `serialise_security_check()` method similarly to what it's already done with the `serialise_sender()` method.

xoen (dismissed):
LGTM 👍 

The module autodiscovery thingy is clever but a bit magical but if it works it's fine by me ☺️

xoen (approved):
LGTM 👍

PR #666, opened 2021-06-08
Export data to Linkspace

ushkarev:
depends on #665


xoen (approved):
LGTM 👍 

Nice work as usual Igor ☺️ - I've only looked at the last few commits (as I'm assuming the rest was same as other base PR)

Let me know if anything change and you want me to have another look ☺️

xoen (approved):
Had another look and it still looks good to me 👍

ushkarev:
No changes; just rebased

requirements/base.txt
```diff
@@ -17,6 +17,7 @@ reportlab>=3.6,<4
 numpy>=1.21,<2
 scipy>=1.7,<2
 openpyxl~=3.0
+pyjwt~=2.3.0
```
xoen, line 20: [CI doesn't seem happy about this](https://app.circleci.com/pipelines/github/ministryofjustice/money-to-prisoners-api/1070/workflows/0a37dd3c-918f-4bf5-80bc-3fde89d1b53f/jobs/2494):

```
ERROR: Cannot install -r requirements/base.txt (line 3) and pyjwt~=2.3.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested pyjwt~=2.3.0
    money-to-prisoners-common 13.3.0 depends on PyJWT~=2.1.0
```

Don't know if it's worth:
1. removing this dependency from here and use same version as for `mtp-common` (I think this is what we do for most packages already required by `mtp-common`)
2. be explicit here but use same version - will break when `mtp-common` change
3. upgrade version in `mtp-common` and leave this - but more fuff than anything, and this would break if we change `mtp-common`'s version again.

Probably easiest is 1st option, just remove `pyjwt` dependency from here 🤷

ushkarev: yeah i actually added a tiny note to the PR msg: this depends on an unreleased version of mtp-common now. mtp-common no longer uses pyjwt so it made sense to remove the dependency there (which dependabot wanted to update anyway). this branch may never get merged, but by that point i think we'll have released common

xoen (approved):

requirements/base.txt
xoen: Oh sorry, you're right, it says in the PR description but I missed it!

xoen (approved):

xoen (approved):

PR #668, opened 2021-06-17
Show error message if it's possible that the wrong prison was chosen when uploading prisoner balances

ushkarev:


mtp_api/apps/prison/views.py
```diff
@@ -196,4 +196,35 @@ def form_valid(self, form):
...
+        # but since the wrong prison can be accidentally chosen, an error should show if this might have happened
+        likely_prison = self.prison_based_on_known_locations(chosen_prison)
+        if likely_prison and likely_prison != chosen_prison:
+            messages.error(self.request, gettext(
+                'File was uploaded for %(chosen_prison)s, '
+                'but database suggests that most prisoners are likely at %(likely_prison)s.'
+            ) % {
+                'chosen_prison': chosen_prison,
+                'likely_prison': likely_prison,
+            } + ' ' + gettext('Was the right prison selected?'))
```
xoen, line 212: Do you reckon is too much to also suggest an action to correct? Or would it be too much text to read 🤷

xoen (approved):
Nice one 👍

mtp_api/apps/prison/views.py
ushkarev: i didn't want it to _force_ someone to do something different. just being a bit cautious… even though i can't think of a scenario where it thinks you chose the wrong prison, but actually you submitted the right one. i suppose it could happen if a balance spreadsheet is really really small.

PR #669, opened 2021-07-08
Performance platform

xoen:
Added functionalities to store, update and retrieve Performance Data which is:
1.  going to be transformed to CSV by `send-money` ([see this other twin PR](https://github.com/ministryofjustice/money-to-prisoners-send-money/pull/422))
2. ultimately be linked from data.gov.uk

The data in the `PerformanceData` table is a weekly aggregation of other data stored in other tables:
- credits total is calculated by counting the credited `Credit` records for the weeks
- digital take-up data is derived from data in `performance.models.DigitalTakeup`
- user satisfaction data is derived from data in `performance.models.UserSatisfaction`

This data is updated using the `update_performance_data` management command (introduced here).

Key principles:
- some of this data may not be available so all the fields are optional
- client should generate final CSV file with minimal computation/manipulation

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1891


mtp_api/apps/performance/views.py
```diff
@@ -444,3 +450,69 @@ def get_ticket_count(self, env_tag, app_tag, start_date, end_date):
...
+
+    permission_classes = (
+        IsAuthenticated, SendMoneyClientIDPermissions,
+    )
+
+    def get_queryset(self):
+        today = timezone.localdate()
+        a_year_ago = today - datetime.timedelta(weeks=52)
+
+        filters = {
```
ushkarev, line 467: if i recall correctly, we're not using django-filters filter sets because there was no way to specify fallback values when none were provided, right?

i know this is an internal-only api but i'm not a huge fan of the very-limited validation of `week__gte` and `week__lt`. it shouldn't be possible to inject security problems, but it'd be easy to cause a harmless 500 error.

i wonder if it'd be possible (unless it's too much hassle!) to add a filter set and have these defaults set on the request params somehow. maybe something like `self.request.query_params.setdefault(…)` overriding a method such as `filter_queryset`..?

mtp_api/apps/performance/tests/test_views.py
```diff
@@ -0,0 +1,122 @@
...
+            }
+        )
+
+    def test_percentages_are_formatted(self):
+        mommy.make(
+            PerformanceData,
+            week=self._monday_n_weeks_ago(4),
+            digital_takeup=0.95,
+            completion_rate=0.8111,
+            user_satisfaction=0.96666666,
```
ushkarev, line 73: i would also try with another model with a `0` and a `None` somewhere. i'd expect `'0%'` and `''` to come out of it

mtp_api/apps/performance/views.py
```diff
@@ -444,3 +450,69 @@ def get_ticket_count(self, env_tag, app_tag, start_date, end_date):
...
+        }
+
+        return PerformanceData.objects.filter(**filters)
+
+    def list(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+        serializer = self.get_serializer(queryset, many=True)
+
+        response = {
+            'headers': self._headers(),
```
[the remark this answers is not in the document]
xoen: No, not really, didn't really have any alternatives 🤷‍♂️ . To be honest I thought "headers" was a good fit given the purpose of the field is to populate the headers columns of the CSV 🤔

mtp_api/apps/performance/views.py
xoen: It's an internal API but I guess it's a fair point. I'm not really familiar with all these filters but I'll play with alternatives to make this more strict.

mtp_api/apps/performance/views.py
```diff
@@ -444,3 +450,69 @@ def get_ticket_count(self, env_tag, app_tag, start_date, end_date):
...
+        '67%'
+
+        This is so that client can use this data without additional processing
+        """
+
+        percentage_field_names = PerformanceData.percentage_field_names
+
+        for record in records:
+            for field_name in percentage_field_names:
+                if record[field_name]:
```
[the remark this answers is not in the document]
xoen: FYI: https://github.com/ministryofjustice/money-to-prisoners-api/pull/669/commits/d1c291f9817956393506d45309bdb495c2bf0228

mtp_api/apps/performance/admin.py
```diff
@@ -23,3 +24,34 @@ class UserSatisfactionAdmin(ModelAdmin):
...
+    @add_short_description(_('Completion rate'))
+    def view_completion_rate(self, obj):
+        return self._as_percentage(obj.completion_rate)
+
+    @add_short_description(_('User satisfaction'))
+    def view_user_satisfaction(self, obj):
+        return self._as_percentage(obj.user_satisfaction)
+
+    def _as_percentage(self, value):
+        if value:
```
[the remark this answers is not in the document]
xoen: https://github.com/ministryofjustice/money-to-prisoners-api/pull/669/commits/d1c291f9817956393506d45309bdb495c2bf0228

mtp_api/apps/performance/tests/test_views.py
xoen: https://github.com/ministryofjustice/money-to-prisoners-api/pull/669/commits/d1c291f9817956393506d45309bdb495c2bf0228

mtp_api/apps/performance/views.py
xoen: Hey Igor, this is what I came up with: https://github.com/ministryofjustice/money-to-prisoners-api/pull/669/commits/8d30b1941da5bc2c811632898977bed7f1ba6f46

I think it's not bad although there may of course another million ways of doing this as usual :)

Hope it makes sense and it's addressing the issue you had in mind, but we can have a chat about this tomorrow.

ushkarev (approved):

xoen:
Cheers Igor!

PR #670, opened 2021-07-28
Update Performance data on User Satisfaction/Digital Take-up upload

xoen:
Background
=========
Performance Data is derived from User Satisfaction data and from Digital Take-up data (among other things).
These weekly records are created/updated every Monday for the week of 2 weeks ago, to allow for late data (these are updated via [`update_performance_data`](https://github.com/ministryofjustice/money-to-prisoners-api/blob/b439932e864d2d29e186736ec7345ba72e2dc928/mtp_api/apps/performance/management/commands/update_performance_data.py) management command).

This changeset generate/update the Performance Data for the relevant week(s) when User Satisfaction/Digital Take-up is uploaded.

User Satisfaction
============
We tend to upload this CSV file on a monthly basis for the month before. 
On upload the completed weeks are updated, e.g. if last date in CSV file is a Sunday that week including it will be updated, if not only the first n-1 weeks will be updated.
The assumption is that:
- these CSV files contain a continuous range of dates
- the first few dates would complete a previously updated week
- the last week may be complete (generate/update) or not (will be generated/updated on next upload)

Digital Take-up
===========
These xls files tend to be be uploaded more often. They include only the Digital Take-up for a single day.
The logic to determine whether or not to update the corresponding Performance data record is therefore different.

The idea here is that:
- if these xls files were uploaded in a timely manner (e.g. few days after) we don't need to generate/update the Performance data as the management command it will eventually run and use the data (Performance data record for that week not present yet)
- if these xls files are updated "late" there will be already a Performance data record for that corresponding week and that will need to be updated.

Ticket: https://dsdmoj.atlassian.net/browse/MTP-1899

### DIFF
--- a/mtp_api/apps/performance/tests/test_views.py
+++ b/mtp_api/apps/performance/tests/test_views.py
@@ -1,4 +1,5 @@
 import datetime
+from io import StringIO
 
 from django.urls import reverse
 from django.utils import timezone
@@ -6,7 +7,8 @@ from model_mommy import mommy
 from rest_framework import status
 from rest_framework.test import APITestCase
 
-from core.tests.utils import make_test_users
+from core.models import ScheduledCommand
+from core.tests.utils import create_super_admin, make_test_users
 from core.utils import monday_of_same_week
 from mtp_auth.tests.utils import AuthTestCaseMixin
 from performance.models import PerformanceData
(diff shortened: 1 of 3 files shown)


mtp_api/apps/performance/tests/test_views.py
```diff
@@ -166,3 +169,88 @@ def _monday_n_weeks_ago(self, weeks_ago):
...
+        self.upload_url = reverse('admin:digital_takeup_upload')
+        self.fixture_path = pathlib.Path(__file__).parent / 'files'
+        self.fixture_path = self.fixture_path / 'Money_to_Prisoner_Stats - 02.11.16.xls'
+
+        self.test_week = datetime.date(2016, 10, 31)  # Monday of week containing 2nd Nov 2016
+
+    def test_updates_performance_data_when_already_there(self):
+        PerformanceData.objects.create(week=self.test_week)
+
+        with open(self.fixture_path, 'rb') as f:
```
ushkarev, line 232: fyi: you can open a `pathlib.Path` object directly with `self.fixture_path.open('rb')`. traditionally, you'd have `self.fixture_path` being a `str` and using the plain `open` builtin to open it. it amounts to the exact same effect, but technically this is a mix of traditional and newer styles.

ushkarev (approved):

mtp_api/apps/performance/tests/test_views.py
xoen: Thanks @ushkarev - admittedly I'm still somewhat a Python standard lib noob so these comments are gold 💯 

I've changed the test and amended that commit, so I'm afraid you'll have to have another look at the last commit/re-approve ☺️

ushkarev (approved):

PR #671, opened 2021-07-30
Update dependencies

ushkarev:
• mtp-common brings along many other library updates, none should have breaking changes
• move testing-only dependencies into appropriate requirements file
• update only the simplest non-testing libraries

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,8 +1,8 @@
 # Dependencies needed for all environments
 
-money-to-prisoners-common~=12.1.0
+money-to-prisoners-common~=12.2.0
 
-psycopg2-binary>=2.8.4,<2.9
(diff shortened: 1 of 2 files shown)


requirements/base.txt
```diff
@@ -1,8 +1,8 @@
 # Dependencies needed for all environments
 
-money-to-prisoners-common~=12.1.0
+money-to-prisoners-common~=12.2.0
 
-psycopg2-binary>=2.8.4,<2.9
+psycopg2-binary~=2.8.6
```
xoen, line 5: 👍  - surprisingly version `2.9` breaks compatibility: https://github.com/psycopg/psycopg2/issues/1293

xoen (approved):
LGTM 👍

PR #672, opened 2021-08-16
Minor tidy

ushkarev:


mtp_api/apps/security/tests/utils.py
```diff
@@ -230,8 +230,8 @@ def generate_checks(
             check = candidate_payment.credit.security_check
             if (not overrides or 'state' not in overrides) and j % 5:
                 check.status = random.choice([CHECK_STATUS.ACCEPTED, CHECK_STATUS.REJECTED])
-                check.actioned_at = datetime.now(tz=pytz.utc)
-                check.actioned_by = random.choice(list(Group.objects.filter(name='FIU').first().user_set.all()))
+                check.actioned_at = timezone.localtime()
```
xoen, line 233: @ushkarev you think there was any reason why the time was UTC instead of local time?

Don't think this should be an issue tho - in fact it is more accurate compared to the UTC time (although they should both represent the same point in time).

ushkarev: `actioned_at` is just a record of the date/time an even happened: it's more natural to express it in the local timezone. but as you say, it doesn't in fact matter because the timezone is preserved and the date if formatted appropriately when displayed. i think i mainly did this to get rid of the `pytz` reference; it looked significant but in reality was not

xoen: ...and it's a test utility thing anyway. I still think use `timezone.localtime()` for everything is better for consistency and to reduce risk of weird time bugs 👌

xoen (approved):
👍

mtp_api/apps/security/tests/utils.py
ushkarev: yeah i agree. i considered changing _all_ `timezone.now()` to `timezone.localtime()` but decided that'd be going too far :)

PR #673, opened 2021-08-19
Replace usage of Mailgun with GOV.UK Notify [MTP-1901]

ushkarev:


mtp_api/templates/mtp_auth/new_user.html
```diff
@@ -1,29 +0,0 @@
...
-      <th valign="top" align="left">{% trans 'Username' %}</th>
-      <td>{{ username }}</td>
-    </tr>
-    <tr>
-      <th valign="top" align="left">{% trans 'Password' %}</th>
-      <td>{{ password }}</td>
-    </tr>
-  </table>
-
-  {% if login_url %}
```
xoen, line 22: Not quite sure how this `login_url` could be empty. Maybe some roles have no login URL?
Or maybe it was just a guard against some edge case?

mtp_api/apps/credit/management/commands/create_prisoner_credit_notices.py
```diff
@@ -97,10 +92,21 @@ def handle(self, path, prison, date=None, **options):
...
-        except (TypeError, KeyError, ValueError, requests.RequestException):
-            return
+
+def parsed_date_or_yesterday(date) -> datetime.date:
+    if date:
+        date = parse_date(date)
+        if not date:
+            raise CommandError('Date cannot be parsed, use YYYY-MM-DD format', {'date_string': date})
+    else:
+        date = now().date() - datetime.timedelta(days=1)
```
xoen, line 102: I know this function's implementation is extracted straight as it was, but wouldn't be better to use the usual `django.utils.timezone.localdate()` instead of `now().date()`?

mtp_api/apps/transaction/utils.py
```diff
@@ -37,9 +37,9 @@ def format_number(value, truncate_after=None):
     """
     if truncate_after is not None and value > truncate_after:
         if value >= 1000000:
-            return '{:,}m'.format(value / 1000000)
+            return '{:0,.1}m'.format(value / 1000000)
         if value >= 1000:
-            return '{:,}k'.format(value / 1000)
+            return '{:0,.1}k'.format(value / 1000)
```
xoen, line 42: I'm always finding these format strings a little bit confusing/terse.

I'm guessing the intent of that `,.1` here is to round to the 1st decimal place.
I've tried in the console and it spits out scientific notation (which would confuse most people 😬), e.g.:

```python
>>> '{:,.1}k'.format(100.244089)
'1e+02k'
```

But it rounds to the 1st decimal when adding an `f` instead, e.g.:

```python
>>> '{:,.1f}k'.format(100.244089)
'100.2k'
```

xoen (commented):
It all looks good to me, I've left a few comments/questions.

Only minor thing that may be not right is in the the format string inside `format_currency_truncated()` but rest looks great ☺️

mtp_api/templates/mtp_auth/new_user.html
ushkarev: It used to be possible for it to be empty, but it's no longer the case (as of a4323f7d2653c4d4a426bdcbe4381eefdd69d84d) but it looks like we didn't remove the unnecessary is-it-blank checks

mtp_api/apps/credit/management/commands/create_prisoner_credit_notices.py
ushkarev: yeah we might as well!

mtp_api/apps/transaction/utils.py
ushkarev: you're quite right! i copied it wrong from the currency thing, but also wasn't aware that it would use scientific notation otherwise

xoen (approved):

xoen (dismissed):

xoen (approved):

PR #689, opened 2021-12-10
XLSX digital take-up reports

ushkarev:
Allows digital take-up reports to be uploaded in XLS and XLSX formats with a variable number of empty gutter rows and columns. Originally, NOMIS reports started from cell A2, now it’s B2.


mtp_api/apps/core/excel.py
```diff
@@ -0,0 +1,141 @@
...
+        raise NotImplementedError
+
+
+class XLSWorksheet(ExcelWorksheet):
+    extension = 'xls'
+
+    def __init__(self, workbook: XLSWorkbook, index: int):
+        self.workbook = workbook
+        self.index = index
+        self._worksheet = self.workbook._workbook.get_sheet(index)
```
xoen, line 118: As discussed offline, I think it's worth playing with the idea of getting the concrete sheet in the `WorkBook`s classes as that would avoid the need of accessing another class "private" properties. And I think it's fair for the Workbook class to do that and pass it to the sheet class.

(Oh I see you're already playing with it ❤️ 👍)

xoen (dismissed):
Nice work Igor, I like the start row/col which makes the parsing less brittle and these wrapper classes makes the client code more readable/high level 👍

xoen (approved):
Nice 👍

PR #733, opened 2022-07-05
Update mtp-common and dependencies and fix broken credit-filtering tests

ushkarev:


mtp_api/apps/credit/tests/test_views/test_credit_list/test_security_credit_list/test_credit_list_amount_pattern.py
```diff
@@ -19,11 +19,6 @@ def test_amount_pattern_filter_endswith(self):
             'amount__endswith': '000',
         })
 
-    def test_amount_pattern_filter_endswith_multiple(self):
```
ushkarev, line 22: this test was broken. `_get_amount_pattern_checker` (via `_test_response_with_filters`) incorrectly only checked against the last value provided. coincidentally, that’s the same behaviour of `django-filter`’s CharFilter. `'000'` in this test was essentially ignored.

mtp_api/apps/credit/tests/test_views/test_credit_list/test_security_credit_list/test_credit_list_amount_pattern.py
```diff
@@ -4,9 +4,9 @@
 
 
 class AmountPatternCreditListTestCase(SecurityCreditListTestCase):
-    def test_exclude_amount_pattern_filter_endswith_multiple(self):
```
ushkarev, line 7: this test was broken. `_get_amount_pattern_checker` (via `_test_response_with_filters`) incorrectly only checked against the last value provided. coincidentally, that’s the same behaviour of `django-filter`’s CharFilter. `'000'` in this test was essentially ignored.

neilmendum (approved):

PR #734, opened 2022-07-26
MTP-1921: configure requests being sent to AppInsights

neilmendum:
https://dsdmoj.atlassian.net/browse/MTP-1921


neilmendum:
Should have used `./manage.py app deploy test api ...` to deploy to test rather than force merge, but it isn't a huge issue because code is dependent on `APPLICATIONINSIGHTS_CONNECTION_STRING`

PR #738, opened 2022-08-03
MTP-1921: Improve Azure Application Insights request tracing and logging

ushkarev:
uWSGI now runs in lazy mode such that each process loads its own complete application rather than forking the master. This now requires the spooler to setup django.

Lazy apps are needed if telemetry is to be sent to Azure Application Insights: the `opencensus` library starts a background thread on launch to transport telemetry to Azure, but if uWSGI loads the application and _then_ forks, this thread is no longer accessible from the new process.

In `prod` environment:
• `mtp` logs of level INFO and higher
• other logs of level WARNING and higher
• 10% of requests traces
…will be sent.

Depends on [common#539](https://github.com/ministryofjustice/money-to-prisoners-common/pull/539).

### DIFF
--- a/mtp_api/settings/base.py
+++ b/mtp_api/settings/base.py
@@ -265,6 +265,13 @@ LOGGING = {
         },
     },
 }
+if APPLICATIONINSIGHTS_CONNECTION_STRING:
+    # Sends messages from `mtp` logger to Azure Application Insights
+    LOGGING['handlers']['azure'] = {
+        'level': 'INFO',
+        'class': 'mtp_common.application_insights.AppInsightsLogHandler',
+    }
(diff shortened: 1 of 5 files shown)


mtp_api/settings/base.py
```diff
@@ -270,6 +265,14 @@ def callback_add_role_name(envelope):
         },
     },
 }
+if APPLICATIONINSIGHTS_CONNECTION_STRING:
+    # Sends messages from `mtp` logger to Azure Application Insights
+    LOGGING['handlers']['azure'] = {
```
neilmendum, line 270: Neat!

ushkarev: probs better to just edit the "default" settings rather than duplicate?

neilmendum (approved):

PR #739, opened 2022-09-12
Add management command to remove all monitors and saved searches for a given user

ushkarev:
This is useful for when an FIU staff member leaves the group

### DIFF
--- a/mtp_api/apps/security/management/commands/bulk_unmonitor.py
+++ b/mtp_api/apps/security/management/commands/bulk_unmonitor.py
@@ -1,0 +1,41 @@
+import textwrap
+
+from django.contrib.auth import get_user_model
+from django.core.management import BaseCommand, CommandError
+
+from security.models import SavedSearch, PrisonerProfile, BankAccount, DebitCardSenderDetails
+
+User = get_user_model()
+
+
+class Command(BaseCommand):
+    """
+    Removes a user's monitored entities and deletes their saved searches
+    """
+    help = textwrap.dedent(__doc__).strip()
+
+    def add_arguments(self, parser):
+        super().add_arguments(parser)
+        parser.add_argument('username')
+        # TODO: add an option to move monitoring to a different user? requires deduplication
+
+    def handle(self, username, **options):
+        try:
+            user = User.objects.get_by_natural_key(username)
+        except User.DoesNotExist:
+            raise CommandError(f'User "{username}" not found')
+        self.stdout.write(f'Removing items monitored by {user.username} ({user.email})')
+
+        self.delete(SavedSearch, 'saved searches', user)
+        self.delete(PrisonerProfile.monitoring_users.through, 'monitored prisoners', user)
+        self.delete(BankAccount.monitoring_users.through, 'monitored bank account senders', user)
+        self.delete(DebitCardSenderDetails.monitoring_users.through, 'monitored debit card senders', user)
+
+    def delete(self, model, description, user):
+        self.stdout.write(f'Deleting all {description}…')
+        count, deleted_models = model.objects.filter(user=user).delete()
+        if count:
+            for model, count in deleted_models.items():
+                self.stdout.write(f'Deleted {count} × {model}')
+        else:
+            self.stdout.write(f'No {description} to delete')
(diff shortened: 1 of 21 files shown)


mtp_api/apps/security/management/commands/bulk_unmonitor.py
```diff
@@ -0,0 +1,45 @@
...
+def find_user(id_or_username):
+    if id_or_username.isdigit():
+        try:
+            return User.objects.get(pk=id_or_username)
+        except User.DoesNotExist:
+            pass
+    try:
+        return User.objects.get_by_natural_key(id_or_username)
+    except User.DoesNotExist:
+        pass
```
xoen, line 45: Potentially you could reuse the same `try`/`except` block for both methods, e.g.

```python
def find_user(id_or_username):
    try:
        if id_or_username.isdigit():
            return User.objects.get(pk=id_or_username)
        else:
            return User.objects.get_by_natural_key(id_or_username)
    except User.DoesNotExist:
        return None
```

mtp_api/apps/security/management/commands/bulk_unmonitor.py
```diff
@@ -0,0 +1,45 @@
...
+        self.delete(PrisonerProfile.monitoring_users.through, 'monitored prisoners', user)
+        self.delete(BankAccount.monitoring_users.through, 'monitored bank account senders', user)
+        self.delete(DebitCardSenderDetails.monitoring_users.through, 'monitored debit card senders', user)
+
+    def delete(self, model, description, user):
+        self.stdout.write(f'Deleting all {description}…')
+        count, deleted_models = model.objects.filter(user=user).delete()
+        if count:
+            for model, count in deleted_models.items():
+                self.stdout.write(f'Deleted {count} × {model}')
```
xoen, line 31: Do we loop through the result of `delete()` because this could potentially delete other associated records and this will be shown here as well?

Maybe this is why this is not equivalent to the simpler?

```python
        count, _ = model.objects.filter(user=user).delete()
        if count:
             self.stdout.write(f'Deleted {count} × {model}')
```

mtp_api/apps/security/management/commands/bulk_unmonitor.py
```diff
@@ -0,0 +1,45 @@
...
+    def handle(self, user, **options):
+        user = find_user(user)
+        if not user:
+            raise CommandError(f'User "{user}" not found')
+        self.stdout.write(f'Removing items monitored by {user.username} ({user.email})')
+
+        self.delete(SavedSearch, 'saved searches', user)
+        self.delete(PrisonerProfile.monitoring_users.through, 'monitored prisoners', user)
+        self.delete(BankAccount.monitoring_users.through, 'monitored bank account senders', user)
+        self.delete(DebitCardSenderDetails.monitoring_users.through, 'monitored debit card senders', user)
```
xoen, line 24: Just to be sure this is doing what I think it's doing: This is deleting the many-to-many records associated with the user, e.g. `BankAccount` <-> `User`, not the `BankAccount` itself right? (I think it's the case but double checking)

xoen (dismissed):

mtp_api/apps/security/management/commands/bulk_unmonitor.py
ushkarev: there's at least 1 username that is only digits so it could never be looked up. this whole thing is overkill, i should've forced the use of usernames *or* ids

mtp_api/apps/security/management/commands/bulk_unmonitor.py
ushkarev: the delete cascades so ends up with several models being deleted (e.g. SavedSearch deletes SavedFilter as well) so there are several lines of output generated

mtp_api/apps/security/management/commands/bulk_unmonitor.py
ushkarev: bingo! django calls it a "through" model which, in all these cases, are automatically generated

mtp_api/apps/security/management/commands/bulk_unmonitor.py
xoen: Oh I see, sorry my bad, it's a fair point, I didn't realise usernames could be only digits...

But it's a good point, maybe it's just simpler if the command accept just a PK or just a username.

My preference would be for the username as they're less error prone and it's easy to accidentally put in a slightly different PK and delete the wrong user by accident. Less likely with usernames plus you have some sort of visual feedback you're actually deleting Alice rather than accidentally delete Bob.

xoen (approved):

PR #746, opened 2023-01-13
Update dependencies

ushkarev:


mtp_api/apps/prison/tests/test_utils.py
```diff
@@ -17,14 +18,14 @@
...
 class LoadPrisonerLocationsFromDevPrisonAPITestCase(TestCase):
-
     fixtures = [
         'initial_types.json',
         'initial_groups.json',
         'dev_prison_api_prisons.json',
     ]
 
     def setUp(self):
+        cache.clear()
```
ushkarev, line 28: other tests might result in the auth token being cached so the POST call to `f'{settings.HMPPS_AUTH_BASE_URL}oauth/token'` is never made

xoen (approved):

PR #756, opened 2023-05-15
MTP-1961: Attempt to upgrade to Django 3.2

xoen:
⚠️ **DO NOT MERGE**: This is a work in progress and depends on a non-released version of `mtp-common`


mtp_api/apps/core/filters.py
```diff
@@ -14,6 +13,7 @@
 import django_filters
 import django_filters.fields
 import django_filters.utils
+import six
```
ushkarev, line 16: line 140 can be turned into this actually: `get_all_format_lazy = lazy(get_all_format, list)`
i can't see why tuple or str needs to be in there

xoen: I don't know much about that so I'll just replace `six.text_type` with `str` in case there is a reason 🤷

ushkarev: `get_all_format` returns a list so i think the other options aren't necessary

xoen:
(rebased)

xoen:
(rebased)

mtp_api/settings/base.py
```diff
@@ -387,3 +387,7 @@
     from .local import *  # noqa
 except ImportError:
     pass
+
+
+# TODO: remove after upgrade to django 3.2
+DEFAULT_HASHING_ALGORITHM = 'sha1'
```
xoen, line 393: How is this used? Does it have an impact on things which are persisted in the DB and make moving to a different hashing method problematic? (e.g. users passwords or something like that)

mtp_api/apps/core/admin.py
```diff
@@ -91,6 +91,7 @@ def has_permission(self, request):
 
 
 site = AdminSite()
+site.enable_nav_sidebar = False
```
xoen, line 94: ✨

mtp_api/settings/base.py
ushkarev: wherever a piece of data is signed (e.g. session cookie), it used to use sha1. the future default is sha256, the docs suggest using this setting _during_ migration if there are both old and new djangos running concurrently (which will be the case for ~5min)

mtp_api/apps/core/admin.py
ushkarev: i thought it looked ugly :p

mtp_api/templates/admin/performance/digitaltakeup/report.html
```diff
@@ -151,14 +151,14 @@ <h3>{{ form.show_predictions.label }} ‡</h3>
               {% regroup rows by predicted as grouped_rows %}
               {% for rows in grouped_rows %}
                 {% if rows.grouper %}
-                  <tr class="{% cycle 'row1' 'row2' as row_styles %}">
+                  <tr>
```
xoen, line 154: Sorry I think I removed most of these but I must have missed these this one 🤔

mtp_api/settings/base.py
xoen: Ah OK, so it sounds like it's only used for sessions which means worst case scenario user is logged out/needs to log back in 👍

xoen:
FYI: New commits so far make sense to me (I can't approve my "own" PR and I'm adding a comment so that I can have an idea of what I've looked at and what not in the future)

mtp_api/apps/service/tests/test_downtime.py
```diff
@@ -4,66 +4,64 @@
...
 from django.utils import timezone
 
-from service.models import Downtime, SERVICES
+from service.constants import Services
+from service.models import Downtime
 
 
 class DowntimeTestCase(TestCase):
-    def _downtime_is_active(self, service, start, end):
-        downtime = Downtime(service=service, start=start, end=end)
```
xoen, line 12: Uh, does it mean this was wrong/broken?

ushkarev: no, it wasn't broken. `<enum instance>.value` is basically the same as `<enum instance>` where a string is expected

xoen: Oh because `TextChoices` is also a string, gotcha! I prefer the `.value` one as it's explicit/more clear

mtp_api/apps/service/views.py
```diff
@@ -2,25 +2,27 @@
...
 
-from service.models import Downtime, SERVICES, Notification
+from service.constants import Service
+from service.models import Downtime, Notification
 from service.serializers import NotificationSerializer
 
 
 def service_availability_view(_):
-    def service_availability(service):
+    def service_availability(service: str):
```
xoen, line 11: I'm not saying change anything as we're already moving a lot of things around but maybe service could be a `Service` here and in `active_downtime()` rather than a String.

I'm just saying this because it took me a few minutes to understand how these methods work and I thought there was a problem but actually this seems fine.

If this was a service everything would be nicer as you're dealing with an higher level concept rather than a string (which could be anything).

e.g.  below you'd have `response = {service: service_availability(service) for service in Service}`

and in `active_downtime()` the query would be more explicit by saying `service`:

```Python
class DowntimeManager(models.Manager):
    def active_downtime(self, service: Service):
        # [...]
        return self.filter(
            # [...]
            service=service.value, # service.value now is explicit
        ).order_by('-end').first()
```

It's just an observation, maybe slightly overkill for this simple methods but probably more meaningful in other places using `TextChoices`.

xoen: Basically by passing around strings you "loose" information about the enum/TextChoice type (Service in this case). Which could lead to bugs.

mtp_api/apps/credit/admin.py
```diff
@@ -51,7 +51,7 @@ class PaymentAdminInline(admin.StackedInline):
     extra = 0
     readonly_fields = ('uuid', 'status', 'batch', 'billing_address',)
 
-    def has_add_permission(self, request):
+    def has_add_permission(self, request, obj=None):
```
xoen, line 54: This didn't really have any affect tho because method just returns `False` right?

mtp_api/templates/admin/base_site.html
```diff
@@ -3,7 +3,7 @@
 
 {% block branding %}
   {% if APP_GIT_COMMIT %}
-    {% blocktrans trimmed with version=APP_GIT_COMMIT|truncatechars:7 asvar version_description %}
+    {% blocktrans asvar version_description trimmed with version=APP_GIT_COMMIT|truncatechars:7 %}
```
xoen, line 6: What's this?

mtp_api/apps/notification/management/commands/send_notification_emails.py
```diff
@@ -16,7 +16,7 @@
 
 class Command(BaseCommand):
     def handle(self, **options):
-        frequency = EMAIL_FREQUENCY.DAILY
+        frequency = EmailFrequency.daily
```
xoen, line 19: Given this is pretty much a constant (AFAICT) you could move it out and call it `DAILY` (or something like that) as it may read a little bit better, e.g.

```Python
period_start, period_end = get_notification_period(DAILY)
# [...]
preferences = EmailNotificationPreferences.objects.filter(frequency=DAILY).exclude(last_sent_at=today)
```

it's fine tho, I know we're not rewriting/tidying up everything as part of this :P

mtp_api/apps/credit/admin.py
ushkarev: overridden methods should have arguments compatible with their parent. dunno _why_ the parent might get an `obj` and i don't think it was hit in practice, but there'd be an error if somehow obj was provided to, but not accepted by, this method

mtp_api/templates/admin/base_site.html
ushkarev: IDE said to do it 🤷

mtp_api/apps/notification/management/commands/send_notification_emails.py
ushkarev: at some point in the past, it was supposed to be configurable. `EmailFrequency` is entirely redundant

mtp_api/apps/notification/utils.py
```diff
@@ -2,25 +2,23 @@
...
+    elif email_frequency == EmailFrequency.weekly:
         start_of_this_week = today - timedelta(days=today.weekday())
         return start_of_this_week - timedelta(days=7), start_of_this_week
-    elif email_frequency == EMAIL_FREQUENCY.MONTHLY:
+    elif email_frequency == EmailFrequency.monthly:
         first = today.replace(day=1)
         return (first - timedelta(days=1)).replace(day=1), first
-    elif email_frequency == EMAIL_FREQUENCY.NEVER:
+    elif email_frequency == EmailFrequency.never:
         return today, today - timedelta(days=1)
```
xoen, line 23: how does this "never" frequency work?

just have end < start would make queries work? I guess it would be the same with `(today, today)` but maybe this is confusing as it sounds like the range is for "today".

mtp_api/apps/notification/management/commands/send_notification_report.py
```diff
@@ -319,13 +317,13 @@ def serialise(self, worksheet, record: Disbursement, triggered: Triggered):
             'Prisoner name': record.prisoner_name,
             'Prison': record.prison.short_name,
             'Recipient name': record.recipient_name,
-            'Payment method': str(payment_method),
+            'Payment method': str(DisbursementMethod[record.method].label),
```
xoen, line 320: I guess the record is in the DB and it _should_(?) only have one of the valid enum values so we can safely use `[]`.

ushkarev: yup!

xoen:
So far so good :)

ushkarev (approved):

PR #759, opened 2023-05-23
Fix Digital take-up report exception when no data

xoen:
This is happening when running the api locally and there is no data. Nevertheless it's good to handle it better as it may confuse someone not familiar with MTP.


xoen:
@ushkarev this still happened when running `api` in `main` against usual version of `mtp-common` so I decided to fix it to avoid future confusion.

PR #776, opened 2023-12-18
MTP-1970: Command to delete data older than 7 years

xoen:
`delete_old_data` (in core app) to delete data older than 7 years:
- deletes old credits (using `created` column)
  - added index (concurrently) to credit_credit table on `created` column
  - deleting a credit will CASCADE delete a bunch of other records, e.g. credit, comments, logs, events, security check, transaction
- deletes old disbursements (using `created` column)
  - deleting a disbursement will CASCADE delete other records like comments, logs, events, etc...
- deletes old transactions (using `received_at` column)
- deletes old payments (using `modified` column)
- deletes old user events (using `timestamp` column)
- deletes old saved searches
- output is indented to try and make the relationship/chain of events that lead to some data being deleted (hopefully this will be useful when testing this command against production-like data and when investigating any potential problems)
- when deletion is performed output with numbers of records deleted is printed out (so that we can see what other linked records were deleted indirectly via CASCADE delete)

The command is more complicated than expected because of the relationship between some of the models:
- Events need to be collected and deleted because `CreditEvent`, `DisbursementEvent`, etc...do not CASCADE delete those and they would otherwise be orphans
- SenderProfile/PrisonerProfile/RecipientProfile needs to be checked individually
  - can be deleted if they have no credits/disbursements left but they need to be updated otherwise (they hold counts with credits/disbursements which are no longer accurate after a credits/disbursements are linked)
  - if they're not deleted at this stage they would be left orphans
  - deletion of these records CASCADE delete a bunch of other records which contains sensitive information and it's important to delete (e.g. addresses, emails, etc...)
  - deleting a sender profile or prisoner profile will also cause the deletion of any related SavedSearches
    - this table is not explicitly linked with SenderProfile/PrisonerProfile (not at DBMS level) but the `site_url` column contains a reference to the these records. Deleting them is important because trying to display a SavedSearch with a broken link would likely cause application errors.
- similarly Bank Accounts need to be checked/deleted as well or they would be left orphans
  - BankAccounts are linked via the many-to-many relationship `BankTransferSenderDetails` and `BankTransferRecipientDetails`, these records are CASCADE deleted when the sender is deleted but the Bank Accounts are not. That's why they need to be collected explicitly.
  - this is particularly important because `BankAccount` don't even have timestamp fields so we wouldn't have an easy way to collect them later
  - BankAccount are checked to see if they have any linked sender/recipient before deleting them
- Billing Addresses are another of these models with "weird" relationships that can lead to orphan records
  - BillingAddresses contains sensitive information and have no timestamp fields
  - A Sender Profile has (potentially) many Billing Addresses via `DebitCardSenderDetails`
  - when a Sender Profile is deleted the `DebitCardSenderDetails` records are CASCADE deleted but the foreign key on the `BillingAddress` record is set to `NULL` (so potentially leaving the Billing Address orphan)
  - A BillingAddress also has many Payments. When a (debit card) Credit the related Payment is deleted which means potentially the BillingAddress is not needed anymore
  - for this reason when deleting a credit these BillingAddresses are collected and then checked to see if they're orphan and they should be deleted (e.g. they don't have a `DebitCardSenderDetails` and no payments linked)

Tests
- Test setup is not straightforward and there are many records and related records, and different states for things that should be deleted or not (some records should be updated, e.g. profiles)
- Test data was created using a fixture (that was fun) and then individual records were loaded in the test setup
- I eventually settled for having some sort of naming convention where records would end with `_deleted` if they are supposed to be deleted (e.g. old) or `_update` if they're supposed to be updated (e.g. profiles counts) or no suffix for records which are mostly left alone
- I tried to create some scenarios similar to what we'd have in production data, however there are so many records and this is practically "impossible" or at least impractical, and it would just make the understanding of the test data more complicated. However I hope the data is representative _enough_ of what we have in production.
- the test includes a marmaid snippet to try and make (some of) the test data clear
  - some of the smaller records are omitted by the main record types (Sender|Prisoner|Recipient Profiles and Credits/Disbursements) are displayed to have a rough idea of the data which is supposed to be deleted or updated. Hopefully this will make the test slightly more understandable.

    ```mermaid
    mindmap
        root
            SenderProfile1_UPDATED
                Credit1_DELETED
                Credit2
                Credit3
            SenderProfile2_DELETED
                Credit4_DELETED
            PrisonerProfile1_UPDATED
                Credit1_DELETED
                Credit2
                Credit3
                Disbursement1_DELETED
                Disbursement3_DELETED
                Disbursement4_DELETED
            PrisonerProfile2_DELETED
                Credit4_DELETED
                Disbursement2_DELETED
            RecipientProfile1_DELETED
                Disbursement1_DELETED
            RecipientProfile2_DELETED
                Disbursement2_DELETED
            RecipientProfile3_UPDATED
                Disbursement3_DELETED
                Disbursement4
    ```

### DIFF
--- a/mtp_api/apps/core/management/commands/delete_old_data.py
+++ b/mtp_api/apps/core/management/commands/delete_old_data.py
@@ -4,6 +4,7 @@ from typing import Sequence
 
 from django.core.management import BaseCommand, CommandError
 from django.db import transaction
+from django.utils import timezone
 
 from core.utils import beginning_of_day, date_argument
 from credit.models import Credit
@@ -26,11 +27,14 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         super().add_arguments(parser)
-        parser.add_argument('--before', help='Delete data before this date (exclusive). Defaults to 7 years ago. Can\'t be later than 7 years ago.')
+        parser.add_argument(
+            '--before',
+            help="Delete data before this date (exclusive). Defaults to 7 years ago. Can't be later than 7 years ago.",
+        )
 
     @classmethod
     def get_cutoff_date(cls, **options):
-        today = beginning_of_day(datetime.date.today())
+        today = beginning_of_day(timezone.localdate())
         seven_years_ago = today - datetime.timedelta(days=7*365)
 
         before = date_argument(options['before'])
--- a/mtp_api/apps/core/utils.py
+++ b/mtp_api/apps/core/utils.py
@@ -17,6 +17,7 @@ def monday_of_same_week(date: datetime.date) -> datetime.date:
     monday = date - datetime.timedelta(days=date.weekday())
     return monday
 
+
 def date_argument(argument) -> datetime.datetime:
     if not argument:
         return None
@@ -24,6 +25,7 @@ def date_argument(argument) -> datetime.datetime:
     if not date:
         raise CommandError('Cannot parse date')
     return beginning_of_day(date)
+
 
 def beginning_of_day(date) -> datetime.datetime:
     return timezone.make_aware(datetime.datetime.combine(date, datetime.time.min))


mtp_api/apps/core/fixtures/test_delete_old_data.json
```diff
@@ -0,0 +1,779 @@
+[
+    {
+        "model": "auth.User",
```
xoen, line 3: I just needed a User because it's required in certain models.

mtp_api/apps/core/fixtures/test_delete_old_data.json
```diff
@@ -0,0 +1,779 @@
...
+            "username": "JOHN42"
+        }
+    },
+    {
+        "model": "security.PrisonerProfile",
+        "pk": 1,
+        "fields": {
+            "prisoner_number": "A1409AE",
+            "prisoner_name": "JAMES HALLS",
+            "credit_count": 2,
```
xoen, line 15: NOTE: `credit_count` only includes `credited` Credits.

mtp_api/apps/core/management/commands/delete_old_data.py
```diff
@@ -0,0 +1,205 @@
...
+        disbursements_to_delete = Disbursement.objects.filter(created__lt=seven_years_ago)
+        for disbursement in disbursements_to_delete:
+            self.delete_disbursement(disbursement)
+
+        self.print_message(f'\nDeleting Transactions older than {seven_years_ago}...')
+        records_deleted = Transaction.objects.filter(received_at__lt=seven_years_ago).delete()
+        self.print_message(f'Records deleted: {records_deleted}.', 1)
+
+        self.print_message(f'\nDeleting Payment older than {seven_years_ago}...')
+        records_deleted = Payment.objects.filter(modified__lt=seven_years_ago).delete()
```
ushkarev, line 53: why does this one use modified date to filter and so should the others use it too instead of created?

ushkarev (approved):

mtp_api/apps/core/management/commands/delete_old_data.py
xoen: Some tables have indexes on different columns, `payment_payment` has an index on `modified` so this takes advantage of that. As discussed some time ago it shouldn't be far off the `created` date anyway (usually minutes/hour as the payment is taken)

mtp_api/apps/core/management/commands/delete_old_data.py
```diff
@@ -0,0 +1,205 @@
...
+    def handle(self, *args, **options):
+        verbosity = options['verbosity']
+
+        self.write = self.stdout.write if verbosity else lambda m: m
+
+        today = timezone.localdate()
+        seven_years_ago = today - datetime.timedelta(days=7*365)
+
+        self.print_message(f'\nDeleting Credits older than {seven_years_ago}...')
+        credits_to_delete = Credit.objects.filter(created__lt=seven_years_ago)
```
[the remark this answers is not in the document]
xoen: there were some warnings but they didn't seem to matter as I was only using dates...having said that, as I added the `--before` param these are all now "aware" datetimes so it should be fine: [: Added before argument to](https://github.com/ministryofjustice/money-to-prisoners-api/pull/776/commits/3f37751573d40d2e4f9e5395623ee73443fa5993)

mtp_api/apps/core/management/commands/delete_old_data.py
```diff
@@ -0,0 +1,205 @@
...
+                self.delete_orphan_bank_accounts(bank_accounts_to_check, depth+2)
+            else:
+                RecipientProfile.objects \
+                    .filter(id=recipient_profile.id) \
+                    .recalculate_totals()
+                self.print_message(f'RecipientProfile {recipient_profile} updated.', depth+1)
+
+        self.update_or_delete_prisoner_profile(disbursement.prisoner_profile, depth+1)
+
+    def update_or_delete_prisoner_profile(self, prisoner_profile: PrisonerProfile | None, depth):
```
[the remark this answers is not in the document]
xoen: 🤔 I kinda liked the idea of having that hidden away in the method, but tbh there are some ifs checking for other profiles "outside" anyway so I'll change.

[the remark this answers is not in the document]
xoen: https://github.com/ministryofjustice/money-to-prisoners-api/pull/776/commits/8391699de7493d7a1c1cd0ad05048a3f6bac7350

ushkarev (approved):

PR #819, opened 2025-02-26
Django 4.2 only

xoen:
Changes needed to upgrade to Django 4.x (currently 3.2 is in use).

NOTE: So far this is _**not**_ using Django 4.x but these are changes required to upgrade (e.g. because something deprecated was removed)


xoen:
Superseded by https://github.com/ministryofjustice/money-to-prisoners-api/pull/821

PR #821, opened 2025-03-18
Upgrade to Django 5.1 (from 3.2)

xoen:
Includes [changes to upgrade to 4.2](https://github.com/ministryofjustice/money-to-prisoners-api/pull/819) and [changes required to upgrade to 5.1](https://github.com/ministryofjustice/money-to-prisoners-api/pull/820).

Seems to run fine locally.


mtp_api/apps/mtp_auth/tests/utils.py
```diff
@@ -29,7 +29,7 @@ def _get_http_authorization_token_for_user(self, user, client_id=None):
             client_id = self.APPLICATION_ID_MAP[group.name]
         application = Application.objects.get(client_id=client_id)
 
-        token = get_random_string()
+        token = get_random_string(length=35)
```
ushkarev, line 32: previous length was 12. given it's just for testing, i suppose it doesn't matter that it's much bigger

mtp_api/apps/performance/tests/test_views.py
```diff
@@ -159,8 +159,8 @@ def test_filtering_invalid_input(self):
 
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         body = response.json()
-        self.assertIn('Enter a valid date', body['week__gte'])
-        self.assertIn('Enter a valid date', body['week__lt'])
+        self.assertIn('Enter a valid date.', body['week__gte'])
+        self.assertIn('Enter a valid date.', body['week__lt'])
```
ushkarev, line 163: same issue as in noms-ops
https://github.com/ministryofjustice/money-to-prisoners-noms-ops/pull/617#discussion_r2007484383

mtp_api/apps/performance/view_dashboard.py
```diff
@@ -132,10 +132,7 @@ def get_next_month(month, year):
 
 
 def make_first_of_month(month, month_year, tz):
-    month_and_year = datetime.datetime(year=month_year, month=month, day=1)
-    month_and_year = tz.localize(month_and_year)
-
-    return month_and_year
+    return datetime.datetime(year=month_year, month=month, day=1, tzinfo=tz)
```
ushkarev, line 135: this won't accidentally roll to the previous day because of daylight savings or something?

mtp_api/apps/transaction/tests/utils.py
```diff
@@ -52,7 +52,7 @@ def random_reference(prisoner_number=None, prisoner_dob=None):
 
 
 def get_noon(dt):
-    return dt.tzinfo.localize(dt.replace(hour=12, minute=0, second=0, microsecond=0, tzinfo=None))
+    return dt.replace(hour=12, minute=0, second=0, microsecond=0)
```
ushkarev, line 55: would this still be noon both during daylight savings and not? i can't tell

mtp_api/apps/mtp_auth/migrations/0021_alter_role_other_groups.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 4.2.20 on 2025-03-13 15:17
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
ushkarev, line 6: did this migration come about cos we didn't `makemigrations` in the last batch of django updates?
cos `auth.0012…` happened in [django 3.1]/9https://docs.djangoproject.com/en/5.1/releases/3.1/#abstractuser-first-name-max-length-increased-to-150)

mtp_api/apps/performance/view_dashboard.py
xoen: I don't think it should roll, It's building a tz aware datetime for the 1st of the month, using the provided tz (presumably `Europe/London`) so it should be midnight in the right time.

I played with this a little and I've given another try and I think this is basically what's happening which makes sense: but as usual I'm happy for you to double check and possibly correct me if I'm wrong here :)

```Python
>>> from zoneinfo import ZoneInfo
>>> london = ZoneInfo('Europe/London')
>>> aug_1 = datetime(2016, 8, 1, tzinfo=london)
>>> str(aug_1) # BST
'2016-08-01 00:00:00+01:00'
>>> feb_1 = datetime(2016, 2, 1, tzinfo=london)
>>> str(feb_1) # GMT
'2016-02-01 00:00:00+00:00'
```

mtp_api/apps/prison/migrations/0024_rename_prisonerbalance_prisoner_number_prison_prison_pris_prisone_e452a0_idx_and_more.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 5.1.7 on 2025-03-13 16:59
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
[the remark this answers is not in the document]
xoen: The rename itself should be relatively quick but Postgres may do other things as a result?

According to the [Postgresql `ALTER INDEX` docs](https://www.postgresql.org/docs/current/sql-alterindex.html):

> RENAME
>
> The RENAME form changes the name of the index. If the index is associated with a table constraint
> (either UNIQUE, PRIMARY KEY, or EXCLUDE), the constraint is renamed as well. There is no effect on
> the stored data.
>
> Renaming an index acquires a SHARE UPDATE EXCLUSIVE lock.

And this is what the [Postgresql explicit locking docs says about `SHARE UPDATE EXCLUSIVE`]:

> SHARE UPDATE EXCLUSIVE (ShareUpdateExclusiveLock)
>
> Conflicts with the SHARE UPDATE EXCLUSIVE, SHARE, SHARE ROW EXCLUSIVE, EXCLUSIVE, and
> ACCESS EXCLUSIVE lock modes. This mode protects a table against concurrent schema changes and
> VACUUM runs.
>
> Acquired by VACUUM (without FULL), ANALYZE, CREATE INDEX CONCURRENTLY, CREATE STATISTICS,
> COMMENT ON, REINDEX CONCURRENTLY, and certain [ALTER INDEX](https://www.postgresql.org/docs/current/sql-alterindex.html) and
> [ALTER TABLE](https://www.postgresql.org/docs/current/sql-altertable.html) variants (for full details see the documentation of these commands).

But I'm not really sure if in practice this means there will be a downtime as a result 🤔

mtp_api/apps/mtp_auth/migrations/0021_alter_role_other_groups.py
xoen: I'm not sure, but possibly we didn't run `makemigrations` before? 🤷‍♂️

mtp_api/apps/mtp_auth/tests/utils.py
xoen: Yeah, I erred on the side of caution and used the biggest length I found across the codebase...even tho, as you say this is just for testing and it shouldn't have any impact on real security or anything.

mtp_api/apps/performance/tests/test_views.py
xoen: FYI: I've bumped mtp-common + [reverted that commit](https://github.com/ministryofjustice/money-to-prisoners-api/pull/821/commits/00df52ed23e9365943009c2abaaea8c9fb1e4549), so this should be sorted as soon as we manage to release mtp-common 19.0.1.

mtp_api/apps/performance/view_dashboard.py
jonesj2193: I may be missing the point here, but does the timezone for this have any relevance? Seems as there is only month and year parsed into the function and it sets the date, does it matter which timezone you are in? My only assumption is that the timezone may be used when reading in the value in some later processes? 
But regarding the original point, I wouldn't have thought daylight savings etc makes a difference seems as there is no time to read into it anyway and a time is arbitrarily set as it is required as part of a datetime datatype.

mtp_api/apps/transaction/tests/utils.py
jonesj2193: From what I can tell looking at it and the implementation in the code, I think what Aldo has said is right. As the datetime variable is just parsed in, the function just replaces those specific values and so the timezone is defined when the original value is set i.e. 
`random_date = timezone.localtime(random_date)`
`noon_random_date = get_noon(random_date)` 
and so should still retain. So I would assume whether it is daylight savings time or not should be dealt with by considering the combination of the date and timezone when being read in further?

mtp_api/apps/performance/view_dashboard.py
xoen: > Seems as there is only month and year parsed into the function and it sets the date, does it matter which timezone you are in?

the time parts are not set so it would be a datetime at midnight...unfortunately midnights can be sneaky when it comes to timezones because even a 1 hour difference can tip the date to the previous date if not careful.

> My only assumption is that the timezone may be used when reading in the value in some later processes?
Yes, these datetimes are used later on to filter DB records (e.g. to get only records within the given month)
Hence why it is important that midnight is midnight and not let's say 23:00 of the previous month or 1:00 of the current month - as that would filter records incorrectly and potentially miss some records (or include ones that they shouldn't be included)

> But regarding the original point, I wouldn't have thought daylight savings etc makes a difference seems as there is no time to read into it anyway and a time is arbitrarily set as it is required as part of a datetime datatype.
When `datetime.datetime()` is called it returns a datetime. In our case the time parts are not passed so it would return a datetime for the given year/month but at midnight).

This function also sets the timezone so that in the summer it's midnight BST (e.g. `00:00:00+01:00`) while in the winter is midnight GMT (e.g. `00:00:00+00:00`) - it knows what to do because tz is something like `Europe/London`.

(hope all of this makes sense Josh)

mtp_api/apps/transaction/tests/utils.py
xoen: > So I would assume whether it is daylight savings time or not should be dealt with by considering the combination of the date and timezone when being read in further?

The previous version was removing the tzinfo of the input date creating a new datetime (naive, hence without timezone information) but then calling `localize()` to get a timezone-aware datetime...

...the new version should do the same, but it only replaces the time parts of the datetime, hence retaining the original datetime tzinfo (e.g. if the datetime was a GMT time, it would be the same date, at midday, still GMT) - see example from console above.

mtp_api/apps/performance/view_dashboard.py
xoen: I'm playing more with the behaviour in Django 3.2 and in the new Django 5.1 to see if I can see any differences.

Here the output on **Django 3.2** of the previous logic (`tz.localize`):

```Python
>>> import datetime
>>> from django.utils import timezone
>>> tz = timezone.get_current_timezone()
>>> tz
<DstTzInfo 'Europe/London' LMT-1 day, 23:59:00 STD>
# August
>>> month_and_year = datetime.datetime(year=2024, month=8, day=1)
>>> month_and_year
datetime.datetime(2024, 8, 1, 0, 0)
>>> month_and_year = tz.localize(month_and_year)
>>> month_and_year
datetime.datetime(2024, 8, 1, 0, 0, tzinfo=<DstTzInfo 'Europe/London' BST+1:00:00 DST>)
>>> str(month_and_year)
'2024-08-01 00:00:00+01:00'
# February
>>> month_and_year = datetime.datetime(year=2024, month=2, day=1)
>>> month_and_year
datetime.datetime(2024, 2, 1, 0, 0)
>>> month_and_year = tz.localize(month_and_year)
>>> month_and_year
datetime.datetime(2024, 2, 1, 0, 0, tzinfo=<DstTzInfo 'Europe/London' GMT0:00:00 STD>)
>>> str(month_and_year)
'2024-02-01 00:00:00+00:00'
```
(note how `timezone.get_current_timezone()` returns a `pytz` time zone object)

And here the output on **Django 5.1** with the new logic (pass `tzinfo` to `datetime()`):

```Python
>>> import datetime
>>> from django.utils import timezone
>>> tz = timezone.get_current_timezone()
>>> tz
zoneinfo.ZoneInfo(key='Europe/London')
# August
>>> month_and_year = datetime.datetime(year=2024, month=8, day=1, tzinfo=tz)
>>> month_and_year
datetime.datetime(2024, 8, 1, 0, 0, tzinfo=zoneinfo.ZoneInfo(key='Europe/London'))
>>> str(month_and_year)
'2024-08-01 00:00:00+01:00'
# February
>>> month_and_year = datetime.datetime(year=2024, month=2, day=1, tzinfo=tz)
>>> month_and_year
datetime.datetime(2024, 2, 1, 0, 0, tzinfo=zoneinfo.ZoneInfo(key='Europe/London'))
>>> str(month_and_year)
'2024-02-01 00:00:00+00:00'
```

The `tzinfo` part is no longer the `pytz`'s time zone but **the net result seems - _to me anyway_ - the same**.

Note that [the Django 4.0 release notes](https://docs.djangoproject.com/en/5.1/releases/4.0/#zoneinfo-default-timezone-implementation) say:

> The move to `zoneinfo` should be largely transparent. Selection of the current timezone, conversion of `datetime` instances to the current timezone in forms and templates, as well as operations on aware datetimes in UTC are unaffected.
>
> However, if you are working with non-UTC time zones, and using the `pytz` `normalize()` and `localize()` APIs, possibly
> with the [TIME_ZONE](https://docs.djangoproject.com/en/5.1/ref/settings/#std-setting-DATABASE-TIME_ZONE)
> setting, you will need to audit your code, since `pytz` and `zoneinfo` are not entirely equivalent.

They say that the move should be largely transparent but also to audit your code. Although I'm not sure what to look for exactly more than what I've done. `pytz`'s time zones are not supposed to be passed as `tzinfo` part when using replace, but AFAICT we're not doing it as we're now dealing with `zoneinfo`'s time zones. Not sure what other differences to look for.

The good news (?) is that this specific function is only used in the dashboard to build queries and these dates are not returned further, propagating to other parts of the system. So unless I'm missing something the "blast radius" of a potential DST bug should be relatively small. But I am more than happy to hear your view and/or if you have any ideas on triple check this further.

mtp_api/apps/transaction/tests/utils.py
xoen: I'm doing more digging, bearing in mind `get_noon()` is only used in tests (`generate_initial_transactions_data()` with random datetimes) and to generate test data (`generate_predetermined_transactions_data()` e.g. `load_test_data` command).

I dropped a `pdb.set_trace()` while running one of the transaction tests (`$ ./run.py test --test-labels mtp_api.apps.transaction.tests.test_views.UpdateRefundTransactionsTestCase.test_patch_refunded`)

This is what I'm getting on **Django 3.2**:

```Python
(Pdb) w
[...]
-> transactions = generate_transactions(transaction_batch=tot)
  .../money-to-prisoners/money-to-prisoners-api/mtp_api/apps/transaction/tests/utils.py(244)generate_transactions()
-> data_list = generate_initial_transactions_data(
> .../money-to-prisoners/money-to-prisoners-api/mtp_api/apps/transaction/tests/utils.py(117)generate_initial_transactions_data()
-> noon_random_date = get_noon(random_date)

# argument `datetime` passed to `get_noon()`
# note the `pytz`'s time zone
(Pdb) random_date
datetime.datetime(2025, 3, 21, 12, 51, 31, tzinfo=<DstTzInfo 'Europe/London' GMT0:00:00 STD>)
# replace time components using `dt.tzinfo.localize()` as in old logic...
(Pdb) dt = random_date
# ...this first remove the `tzinfo` to make it a naive datetime
# then use `localize()` on the input `datetime`'s `pytz` time zone
# to make it an aware date again
(Pdb) manual = dt.tzinfo.localize(dt.replace(hour=12, minute=0, second=0, microsecond=0, tzinfo=None))
(Pdb) manual
datetime.datetime(2025, 3, 21, 12, 0, tzinfo=<DstTzInfo 'Europe/London' GMT0:00:00 STD>)
(Pdb) str(manual)
'2025-03-21 12:00:00+00:00'
```

...and here on **Django 5.1** using replace as in the new logic:

```Python
(Pdb) w
[...]
-> transactions = generate_transactions(transaction_batch=tot)
.../money-to-prisoners/money-to-prisoners-api/mtp_api/apps/transaction/tests/utils.py(244)generate_transactions()
-> data_list = generate_initial_transactions_data(
> .../money-to-prisoners/money-to-prisoners-api/mtp_api/apps/transaction/tests/utils.py(117)generate_initial_transactions_data()
-> noon_random_date = get_noon(random_date)

# argument `datetime` passed to `get_noon()`
(Pdb) random_date
datetime.datetime(2025, 3, 23, 21, 36, 48, tzinfo=zoneinfo.ZoneInfo(key='Europe/London'))
# replace time components as in new logic
(Pdb) manual = random_date.replace(hour=12, minute=0, second=0, microsecond=0)
(Pdb) manual
datetime.datetime(2025, 3, 23, 12, 0, tzinfo=zoneinfo.ZoneInfo(key='Europe/London'))
(Pdb) str(manual)
'2025-03-23 12:00:00+00:00'

# similar story if I do that on a manually built August datetime
(Pdb) in_august = datetime.datetime(2025, 8, 23, 21, 36, 48, tzinfo=zoneinfo.ZoneInfo(key='Europe/London'))
(Pdb) in_august
datetime.datetime(2025, 8, 23, 21, 36, 48, tzinfo=zoneinfo.ZoneInfo(key='Europe/London'))
(Pdb) manual = in_august.replace(hour=12, minute=0, second=0, microsecond=0)
(Pdb) manual
datetime.datetime(2025, 8, 23, 12, 0, tzinfo=zoneinfo.ZoneInfo(key='Europe/London'))
```

As mentioned before, `replace()` just replaces the given components without doing anything clever with the rest (AFAICT). I may of course be wrong, but it seems to be doing the right things. Maybe considering this is a test utility function the potential risk is relatively low? 🤔

ushkarev (approved):

